PR #1, opened 2015-03-02
Moving the code into the repo

calevans:


bmackinney:
@mikevanwinkle please CR solr-for-wordpress-on-pantheon.php and 
solr-options-page.php - the rest of the files are dependencies from composer

solr-for-wordpress-on-pantheon.php
```diff
@@ -0,0 +1,1732 @@
...
+ *
+ * @TODO refactor as an object
+ * 
+ */
+
+require_once(dirname(__FILE__) . '/vendor/autoload.php');
+register_activation_hook( __FILE__, 's4wp_activate' );
+
+function s4wp_submit_schema()
+{
```
mikevanwinkle: I like this style and I think it's PSR compliant but WordPress's styleguide says to but the opening brace on the same line. Not a blocker

calevans: Good point.

Before I submit to WordPress.org I will review both file for compliance with the Style Guide.

solr-for-wordpress-on-pantheon.php
```diff
@@ -0,0 +1,1732 @@
...
+    // Check to see if we have  environment variables. If not, bail. If so, create the initial options.
+
+    if ($errMessage=s4wp_sanity_check()) {
+         wp_die($errMessage);
+    }
+
+    if ($errorMessage = s4wp_submit_schema()) {
+        wp_die('Submitting the schema failed with the message ' . $errorMessage);
+    }
+    $options = s4wp_initalize_options();
```
mikevanwinkle: Will this wipe out options if the plugin is deactivated and then reactivated? Should we check for existing options first?

solr-for-wordpress-on-pantheon.php
```diff
@@ -0,0 +1,1732 @@
...
+}
+
+
+function s4wp_sanity_check()
+{
+    $returnValue = '';
+    $wp_version  = get_bloginfo('version');
+
+    if (getenv('PANTHEON_INDEX_HOST')===false) {
+        $returnValue = __( 'Before you can activate this plugin, you must first activate Solr in your Pantheon Dashboard.', 'solr-for-wordpress-on-pantheon' );
```
mikevanwinkle: We should use the wp_error class here so that up above we can check is_wp_error. That will also automatically integrate with debug bar I think

solr-for-wordpress-on-pantheon.php
```diff
@@ -0,0 +1,1732 @@
...
+    $path        = s4wp_compute_path();
+    $schema      = dirname(__FILE__) . '/schema.xml';
+    $url         = 'https://'. getenv('PANTHEON_INDEX_HOST') . ':' .getenv('PANTHEON_INDEX_PORT') . '/' . $path;    
+    $client_cert = realpath(ABSPATH.'../certs/binding.pem');
+
+    /*
+     * A couple of quick checks to make sure eveyrthing seems sane
+     */
+    if ($errorMessage=s4wp_sanity_check()) {
+        return $errorMessage;
```
mikevanwinkle: See below about using the WP_Error class.

solr-for-wordpress-on-pantheon.php
```diff
@@ -0,0 +1,1732 @@
...
+
+    apply_filters('s4wp_connection_options',$solarium_config);
+
+
+    # double check everything has been set
+    if (!($solarium_config['endpoint']['localhost']['host'] and 
+          $solarium_config['endpoint']['localhost']['port'] and 
+          $solarium_config['endpoint']['localhost']['path'])) {
+        syslog(LOG_ERR, "host, port or path are empty, host:$host, port:$port, path:$path");
+        return NULL;
```
mikevanwinkle: Should you return a WP_Error here?

solr-for-wordpress-on-pantheon.php
```diff
@@ -0,0 +1,1732 @@
...
+            }
+        }
+    } else {
+        // this will fire during blog sign up on multisite, not sure why
+        _e('Post Information is NULL', 'solr4wp');
+    }
+    return $doc;
+}
+
+function s4wp_format_date($thedate) {
```
mikevanwinkle: So I feel like we should use php's datetime object functions here instead of regex ... seems kinda dirty like this. Was this ported from the source plugin?

calevans: Almost ALL of this was ported over from the original plugin. I changed as little as possible to make it work. I can look at this but since I do not know exactly how Solr works, I am hesitant to change things like this until we get the first version out there and working. (MVP and all that stuff) :)

mikevanwinkle: Haha fair enough. I feel the same way.

mikevanwinkle:
Let's just go ahead and merge and I'll finish code review later, there are so many files I'm having trouble viewing the changes in github

calevans:
kk. but really, there are only 2 files you need to review. I would assume
we accept anythig we pull in from composer as solid.

Still a lot of code, I know. :)

Cheers!
=C=

On Thu, Mar 5, 2015 at 4:20 PM, Michael Van Winkle <notifications@github.com
> wrote:

> Let's just go ahead and merge and I'll finish code review later, there are
> so many files I'm having trouble viewing the changes in github
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pantheon-systems/solr-for-wordpress-on-pantheon/pull/1#issuecomment-77464803>
> .
>



-- 
*Culture of Respect <http://bit.ly/1tOIyjG>*
How to find, hire, and retain developers

PR #10, opened 2015-04-02
Make generic

calevans:
I've added back in the host, port, and path fields and added schema. Also split off the pantheon specific parameters into their own init.


joshkoenig:
Merged substantive changes via Erik. Going to take a fresh run at the readme.

PR #11, opened 2015-04-29
Major update to WP Pantheon Solr plug-in.

erik-pantheon:
- The loading of posts has been updated to include the posts other than "pages" and "posts". Admin interface updated to give users the ability to load specific post types.
- Use of AJAX to load posts in separate batches of 500 fixed.
- Add hoc simple query interface added to Admin interface.
- Redundant functionality removed from code.


joshkoenig:
Is anything still blocking a merge here?

PR #36, opened 2015-11-23
Feature/unit tests

allan23:
Initial unit tests, killing of the widget (since we have the built-in WP search), multisite compatibility tweaks, and link to Pantheon docs on activation error. Also implements TravisCI. (Go to http://travis-ci.org/ to activate for repo.)


includes/class-solrpower.php
```diff
@@ -51,7 +51,7 @@ function sanity_check() {
 		$wp_version	 = get_bloginfo( 'version' );
 
 		if ( getenv( 'PANTHEON_INDEX_HOST' ) === false ) {
```
joshkoenig: Could we amend this to make sure we are on Pantheon as well? The goal of this plugin is to be usable on independent Solr infrastructure too.

Something like:

```php
if ( getenv( 'PANTHEON_ENVIRONMENT' ) !== false && getenv( 'PANTHEON_INDEX_HOST' ) === false ) {
```

joshkoenig:
Very cool!

allan23:
@joshkoenig I've updated the if statement in the sanity check.

joshkoenig:
Awesome. Thanks for getting these in here. Happily merging!

PR #42, opened 2016-01-05
Schema.xml Admin Notice

allan23:
Creates admin notice on Pantheon if schema hasn't been submitted or if schema version is newer than what was submitted. Also checks to see if the environment (e.g. dev to test) changed since schema was submitted.

### DIFF
--- a/includes/class-solrpower.php
+++ b/includes/class-solrpower.php
@@ -9,21 +9,36 @@ class SolrPower {
 	private static $instance = false;
 
 	/**
+	 * Array of admin notices.
+	 * @var array 
+	 */
+	var $notices = array();
+
+	/**
 	 * Grab instance of object.
 	 * @return SolrPower
 	 */
 	public static function get_instance() {
 		if ( !self::$instance ) {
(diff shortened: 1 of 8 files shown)


includes/class-solrpower.php
```diff
@@ -46,12 +61,52 @@ function activate() {
...
+	/**
+	 * Inform user is schema.xml needs to be submitted to Apache Solr.
+	 * 
+	 */
+	function schema_notice() {
+
+		$schema_submitted	 = get_option( 'solr_schema' );
+		$saved_path			 = get_option( 'solr_path' );
+		$is_custom			 = get_option( 'solr_custom' );
+		// If the saved environment path is different than what it is now,
```
joshkoenig: I'm not sure about the virtue of saving the environment into the DB, or saving "did we submit?". 

We should ideally be checking the response from the API, which should be recognizable if no schema is found. I'll double-check this.

allan23:
Will rework this.

allan23:
Closing this PR in favor of #57 instead.

PR #57, opened 2016-02-10
Auto submission of schema.xml

allan23:
This will check to see if environment has changed and will submit the schema if it had.

### DIFF
--- a/includes/class-solrpower-api.php
+++ b/includes/class-solrpower-api.php
@@ -10,23 +10,28 @@ class SolrPower_Api {
 
 	/**
 	 * Logging for debugging.
-	 * @var array 
+	 * @var array
 	 */
 	var $log = array();
+
+	var $solr = null;
+
+	var $last_result;
 
 	/**
 	 * Grab instance of object.
 	 * @return SolrPower_Api
 	 */
 	public static function get_instance() {
-		if ( !self::$instance ) {
+		if ( ! self::$instance ) {
 			self::$instance = new self();
 		}
+
 		return self::$instance;
 	}
 
 	function __construct() {
-		
+
 	}
 
 	function submit_schema() {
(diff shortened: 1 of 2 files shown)


bmackinney:
@rachelwhitton please test

includes/class-solrpower.php
```diff
@@ -163,6 +166,29 @@ function default_head() {
 		}
 	}
 
+	/**
+	 * Checks to see if Pantheon environment has changed.
+	 */
+	function environment_check() {
+		$environment = get_option( 'pantheon_env' );
+		// If the environment is the same, let's bail.
+		if ( getenv( 'PANTHEON_ENVIRONMENT' ) === $environment ) {
```
joshkoenig: I don't think this will work as intended because it will re-submit the schema whenever a database is synchronized between different environments. 

What might be a better option is to store an array of environments to which the schema has been auto-posted. This would mean that when the code is deployed to a net-new environment (e.g. first time to Live, or when a new Multidev env is spun up) it will be posted, but not re-posted every time a DB is synced from Live to elsewhere.

joshkoenig:
Probably the best way to do this would be to identify an API endpoint/response signature that can definitively tell us if a schema is present or not. This should be doable from the Pantheon side. I will follow up on this.

allan23:
Just an update on this. Trying to base whether to submit schema on the Solarium HTTPException from the ping. However, the exception does not provide a HTTP response status code.

joshkoenig:
When I `var_dump($e)` on the caught exception I get a nice big object like so:

https://gist.github.com/joshkoenig/ec969cd60605b638ac9c26a4f38f2b38

The second value is the `code` which is 404 in this case. Does that satisfy the need?

allan23:
Interesting. I'm getting a code with the value of ``0`` locally (without schema). I'll have to do some testing on Pantheon servers to see if there's a difference.

joshkoenig:
@allan23 you can only test this on the platform. The whole schema posting mechanism is only possible within our infrastructure.

allan23:
@joshkoenig @ataylorme : The latest commit will check the exception code (thanks Josh!) and then automatically submits the schema if it encounters a 404.

Let me know what you think.

It appears that TravisCI is having issues downloading and installing Solr now, so that is why the builds are failing.

joshkoenig:
@allan23 I can see how the changes are supposed to work and it makes sense to me, though all the tests now appear broken. That will need to be fixed of course.

Also, can we not set a longer transient on success?

allan23:
@joshkoenig Sorry about that. Fixed in latest commit -- also transient expiration increased to five minutes.

joshkoenig:
Awesome. Thanks Allan! 

@ataylorme I'd love to get your review as well?

ataylorme:
@joshkoenig - Yup, I blocked some time on my calendar tomorrow to review.

includes/class-solrpower-api.php
```diff
@@ -342,6 +369,36 @@ private function fetch_stat( $type ) {
...
+			if ( getenv( 'PANTHEON_ENVIRONMENT' ) !== false ) {
+				// Ping Solr.
+				$this->ping_server();
+
+				if ( 404 === $this->last_code ) { // Schema is missing on Pantheon.
+					$schemaSubmit = $this->submit_schema();
+					if ( strpos( $schemaSubmit, 'Error' ) ) {
+						echo '<div class="notice notice-error"><p>';
+						echo '<h2>Solr Power Error:</h2>';
+						echo 'Error submitting schema to ApacheSolr.';
```
ataylorme: @allan23 - Can we add some more details to this message? Advise on a resolution or manual steps. Maybe check `$_ENV['PANTHEON_ENVIRONMENT']` so the user knows where to make a fix.

ataylorme:
Besides my comment above, which is a nitpick, this is looking really good. Nice work @allan23 and props to @joshkoenig for the notes and Gist.

includes/class-solrpower-api.php
joshkoenig: How about this text:

`Error posting schema.xml to ApacheSolr backend, which will prevent content from being indexed. You can try navigating to the admin section to try posting the schema directly. If this problem persists, contact your Solr platform provider.`

allan23: Since this is only Pantheon-specific would the user know who to contact or do you think that message makes sense?

allan23:
@joshkoenig @ataylorme The schema failure message has been updated.

includes/class-solrpower-api.php
ataylorme: @allan23 Let's check for `$_ENV['PANTHEON_ENVIRONMENT']` and, if it exists, customize the message to:
```
Error posting schema.xml to ApacheSolr backend, which will prevent content from being indexed. You can try navigating to the Solr Power admin section in the WordPress dashboard to try posting the schema directly. If this problem persists, open a support ticket from you Pantheon site dashboard.
```

This provides some clarification that the admin section is in the WordPress dashboard and the user should open a support ticket from their Pantheon dashboard.

ataylorme:
> The easiest way to test this will be to set up a new multidev env with a clone of the DB from dev. This will have the plugin enabled (since the DB is copied over) but no schema posted, yet.

Tested the `schema.xml` auto posting again this morning from a multidev and it is still looking good.

includes/class-solrpower-api.php
allan23: @ataylorme Since this portion only runs on Pantheon servers, is the $_ENV check redundant? I've update the message.

PR #61, opened 2016-03-04
Filters for AJAX and admin integration. fixes #59

### LINKED ISSUE #59: We shouldn't block on AJAX/Admin use-cases
Right now we bail if the request is in the admin area of is being done via AJAX. However, if we want to support use-cases like autocomplete or general purpose WP_Query() overriding, we probably can't continue to do that:

https://github.com/pantheon-systems/solr-power/blame/master/includes/class-solrpower-wp-query.php#L29

allan23:
Developers can use the `solr_allow_ajax` and the `solr_allow_admin` filters to allow Solr to work with AJAX and within the wp-admin area.  Fixes #59

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -27,7 +27,13 @@ class SolrPower_WP_Query {
 
 	function __construct() {
 		// We don't want to do a Solr query if we're doing AJAX or in the admin area.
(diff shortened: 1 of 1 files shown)


ericmann:
I've verified this against my own installation and it does allow us to use admin/ajax with Solr as expected.

joshkoenig:
+1 for this change for now, though I'd like to understand what the use-cases are for preventing solr in ajax or admin areas. 

Also we should probably start documenting our available filters and settings. I'll make a separate issue for this.

PR #67, opened 2016-04-19
Add is_search check to found_posts_query filter

yvickers-rci:
This limits clearing of found posts query to solr searching.  Always returning blank breaks paging within wp install on non-search pages.

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -121,7 +121,9 @@ class SolrPower_WP_Query {
 	}
 
 	function found_posts_query( $sql, $query ) {
-
+		if ( !$query->is_search() ) {
+			return $sql;
+		}
 		return '';
 	}
 


joshkoenig:
Thanks for the Pull Request. Can you explain in a bit more detail how we can replicate the issue you are solving with this change?

yvickers-rci:
Yes, on other pages of our site where paging is used such as archive pages, blanking the found posts query causes WP to pull the ID of a post from the original query as the found posts count.  So we were ending up with a post count of over 20,000 (our post id was 20848), which made a lot of empty pages.

joshkoenig:
Ok this is definitely a blocker for the next release. Thanks for the Pull Request!

allan23:
@joshkoenig Just tested this and it resolves the issue with pagination.

Great catch @yvickers-rci!

joshkoenig:
+1 Merging now.

PR #69, opened 2016-04-21
Fix exception handling in ping_server()

joshkoenig:
it looks like we had the wrong exception address here, so we weren't catching failed pings.

This is a good way to potentially resolve #57 - we should have the initial admin page for the plugin do a ping, and if the result is a 404 _and_ we're on Pantheon, that's a clear signal that the schema needs to be posted. 

@allan23 - I think fixing the exception handler is a good idea regardless, and we probably want to at least log something if this happens in general. Let me know what you think about the above idea. If it's not crazy, then maybe we can get that in for the 0.3 release?

### DIFF
--- a/includes/class-solrpower-api.php
+++ b/includes/class-solrpower-api.php
@@ -114,11 +114,11 @@ class SolrPower_Api {
 		try {
 			$solr->ping( $solr->createPing() );
 			return true;
-		} catch ( Solarium\Exception $e ) {
+		} catch ( Solarium\Exception\HttpException $e ) {
+		  	// TODO: log this and or otherwise react.
 			return false;
 		}
 	}
-
 	/**
 	 * Connect to the solr service
 	 * @return solr service object


allan23:
@joshkoenig Just submitted PR #70 which will perform a ping on the admin page (also updated exception).

I like the idea of basing the schema submission upon ping. However, it will require the user to visit the admin page after environment change. The other caveat is if for some weird reason that the Solr server goes down, the schema would get submitted due to a failed ping.

joshkoenig:
Looks like this needs some love to be mergable.

joshkoenig:
@allan23 feel free to merge this once you have the conflicts resolved. At that point we can cut a 0.3!

allan23:
@joshkoenig Actually #70 already includes this code. Probably don't need this PR any more.

joshkoenig:
Durr that's why it won't merge. 📦 

I think we're all clear for a 0.3 release then!

Issue #40, opened 2015-12-15
Improved status on Admin screen


joshkoenig:
We should provide an improved status page as part of the first screen of admin UI. 

- It should do a ping and report the status (which will surface if something is wrong right away)
- We can also show the current count of indexed posts, comments, pages, etc.

PR #70, opened 2016-04-21
Indexing Stats on Settings Page

allan23:
This code will perform a ping on the admin page and also display index stats based upon post type.

Addresses the ping in #69, and indexing status for #40.

### DIFF
--- a/includes/class-solrpower-api.php
+++ b/includes/class-solrpower-api.php
@@ -114,7 +114,7 @@ class SolrPower_Api {
 		try {
 			$solr->ping( $solr->createPing() );
 			return true;
-		} catch ( Solarium\Exception $e ) {
+		} catch ( Solarium\Exception\HttpException $e ) {
 			return false;
 		}
 	}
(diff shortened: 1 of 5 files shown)


joshkoenig:
Agreed that having them visit admin is not a complete solution, but better than nothing. A more generic 404 handler might be possible. Catching the 404 status is specific to the "index not found". The server being down/offline would give connection failures of 5xx status codes. 

However, I'm not following how this PR works. Can we review?

allan23:
Hey @joshkoenig sorry for the confusion. This PR is to address #40, which is status and statistics on the admin page.

I may have jumped the gun a bit by saying it could be a solution to the schema issue; so please ignore that thought because that is not accurate.

I get what you are saying regarding status codes -- where 404 means the schema isn't there and 500 codes mean that there's a server issue. I'll test this out and see if it can resolve #57.

joshkoenig:
Aha got it.

Ok yeah, pull in the exception handler now, and fix the stats. +1

PR #74, opened 2016-06-07
check get_solr result in ping_server

tekezo:
We cannot open the solr-power configuration page in wp-admin because of a fatal error in `ping_server`.

```
PHP Fatal error:  Call to a member function ping() on a non-object
```

`get_solr()` return `NULL` until solr server configuration has been done.
So we should check the result value.

### DIFF
--- a/includes/class-solrpower-api.php
+++ b/includes/class-solrpower-api.php
@@ -111,6 +111,9 @@ class SolrPower_Api {
 	 */
 	function ping_server() {
 		$solr = get_solr();
+		if (!$solr) {
+			return false;
+		}
 		try {
 			$solr->ping( $solr->createPing() );
 			return true;


ataylorme:
@tekezo is this happening when you activate the plugin before enabling/configuring the Solr server?

tekezo:
Yes.
I faced this issue before I set `PANTHEON_INDEX_HOST` and `PANTHEON_INDEX_PORT` environment values.

ataylorme:
@tekezo,
Sorry for the delayed follow up here - I was out at WordCamp Europe.

Currently the Solr Power plugin detects the PHP version and if it is less than 5.4 won't allow the plugin to activate and displays an error in the WordPress dashboard.

I think we can take a similar approach here, not allowing the plugin to be activated and displaying an error in the WordPress dashboard if the `PANTHEON_INDEX_HOST` and `PANTHEON_INDEX_PORT` environment variables are not defined.

The message could be:

> The Solr Power plugin requires `PANTHEON_INDEX_HOST` and `PANTHEON_INDEX_PORT` environment variables to be defined.
> Please define both of these variables in your `wp-config.php` file and re-activate the Solr Power plugin.

As a user would that be clear to you?

tekezo:
@ataylorme Your message approach is awesome! It's clear for me.

ataylorme:
@tekezo I'm closing this in favor of PR #81 which includes the warning message if the environment variables are missing.

Can you test that code and let me know how it works for you? I tested locally but another test would be great.

Thanks!

Issue #37, opened 2015-11-24
Implement basic faceting widget


joshkoenig:
One of the best features of a Solr backend is the ability to easily "facet" the search query. We should make it easy for users to discover this power. Using the built-in categories to power a basic "Faceted Search" widget would go a long way towards this.

Other primitive facets to consider:

- Post type
- Author
- Date (month?)

danielbachhuber:
> I'm not sure of the most "WordPress way" to do this.

Maybe a widget and an easy function to call to render the equivalent HTML.

joshkoenig:
Another thing to consider - should we have a goal of "playing nicely" with facet-based data plugins like [FacetWP](http://facetwp.com)?

akhillgopinath:
Hi 
Is there any way we can show the facets  in search result page ?

ataylorme:
Facets were mentioned in the WordPress.org support forum as well ([link](https://wordpress.org/support/topic/range-facets?replies=1)).

yumyo:
Hi,
any update on this? 

Also, +1 for "playing nicely" with FacetWP

ataylorme:
Hi @yumyo,
This is on our radar for the next release (0.4) but we don't have a definite timeline. You can keep an eye on the [0.4 milestone](https://github.com/pantheon-systems/solr-power/milestones/0.4).

If you would like to help contribute pull requests are welcome!

ataylorme:
Just a quick update - we released 0.4 as auto-submitting schema was a feature worthy of it's own release. Facets are still on the radar and will likely be in the next release or two.

ataylorme:
Closed by #118

JamesJosephFinn:
@joshkoenig you mentioned above

>playing nicely with facetWP

Was that ever woven in here? Thanks!!

ataylorme:
facetWP support didn't make it into the search widget

JamesJosephFinn:
Roger. Thanks.

PR #78, opened 2016-07-01
Basic facet widget implementation.

allan23:
Per issue #37, this PR adds a search widget with facet options.

Currently any facet set on the Solr Options page will display. You can also set custom_fields to be facets as well. (As long as they are specified to be indexed and as a facet.)

Filters are also available so that developers can alter the display of available facets.


pantheon-circleci:
@allan23, thanks for your PR! By analyzing the blame information on this pull request, we identified @joshkoenig, @yvickers and @ataylorme to be potential reviewers

ataylorme:
@allan23 This is totally on the right track and looks really well done.

I'm seeing some unexpected behavior and think we can improve some user experience items.

1) After searching with a facet, such as post type, there is no way to replace the facet with another selection in the search results page (screenshots attached).

solr-facet-initial-search

solr-facet-after-search

When a facet is selected the other available selections need to show up along with the results so a user can change between available items.

2) Let's add a reset button to each facet applied as well as a button to remove all facets from the search.

3) Facets don't show up until after an initial search is performed. An end user might not know they can perform an advanced search if they don't try the simple search first. What do you think of creating an "advanced" accordion/drop down listing all facets?

allan23:
@ataylorme Default facets are available first thing, reset links added below facet items, and filters to add custom fields to index and facets.

ataylorme:
> Default facets are available first thing, reset links added below facet items, and filters to add custom fields to index and facets.

@allan23 this is looking really nice - good work! The last comment I have is that when the text of a search changes and a user hits "Search" again the facets should be reset.

ataylorme:
@allan23 Can you please rebase the master branch of the main repo with your branch for this PR?

ataylorme:
@allan23 can you make this a branch of the Solr Power repo instead of your fork `allan23:issue/37`? The Terminus machine token environment variable in CircleCI isn't passed to forks causing the build to fail.

allan23:
@ataylorme do you want me to do this moving forward or just for this issue? 
I can see how this could be problematic for community contributions.

ataylorme:
@allan23 as we discussed on our call today there seems to be a bug when multiple facets are selected, and shows in the URL, only a single facet is displayed in the widget and search results.

allan23:
@ataylorme This last build is intentionally set to fail. It includes a unit test outlining the issue.
You can select two facets of the same type (two tags for example), but when you try to select two of different types (one tag and one category) no results are found.  
The unit test uses a custom field and a category.

ataylorme:
@allan23 I haven't used Solarium before so I'll dig into this more but could we make use of it's [Facet multiquery feature](http://solarium.readthedocs.io/en/stable/queries/select-query/building-a-select-query/components/facetset-component/facet-multiquery/)?

allan23:
@ataylorme That's the next step I was going to take (in fact I have that documentation bookmarked).

ataylorme:
@allan23 [This article](http://yonik.com/multi-select-faceting/) is pretty interesting. Solr supports the idea of `excludeTags` which allow you to get counts and other info on tags without including them in results. I need to dive deeper to see if Solarium supports this. In the meantime Facet multiquery looks like a good place to start.

ataylorme:
@allan23 [this article](https://enrise.com/2012/11/multi-select-faceting-in-solr-with-solarium/#gist4027542) is good as well. It shows an example of using Solarium's `addFilterQuery` with outer tags.

If I get some time today and tomorrow I'll play around with it. I think this is what we want. Currently `includes/solrpower-api.php` is using `setMinCount` which filters out all elements with zero results. Using `addFilterQuery` with `'tag'=>'outer'` should allow us to return counts for unselected facets without including them in the results.

What do you think?

allan23:
@ataylorme I was looking at the wrong place. It turns out the Filter Query needed an operator. I'll create a new branch so that circleci can run some tests.

ataylorme:
@allan23 closing this in favor of #118, which is a cleaner PR and is where development is happening.

Issue #86, opened 2016-07-15
Document local setup steps


danielbachhuber:
For users who want to work with the plugin locally, we should have a short "how-to" guide for getting Solr running locally.

jlambe:
👍

danielbachhuber:
Just so it's documented, I tried setting this up locally by connecting to a Pantheon Solr instance. However, it appears the Pantheon Solr instance is behind a firewall, and doesn't respond to pings from an external network.

ataylorme:
Merged in #92

PR #92, opened 2016-07-18
First pass at local install instructions

danielbachhuber:
See #86


pantheon-circleci:
@danielbachhuber, thanks for your PR! By analyzing the blame information on this pull request, we identified @rachelwhitton, @allan23 and @joshkoenig to be potential reviewers

danielbachhuber:
@ataylorme Thoughts on this? What else could we include?

allan23:
@danielbachhuber Defining environmental variables would be good to add as well:

``putenv('PANTHEON_INDEX_HOST=192.168.50.4');``
``putenv('PANTHEON_INDEX_PORT=8983');``

I use VVV for local dev and just add this to my wp-config.php.

I also add this as a plugin too:

``add_filter('solr_scheme',function(){
return 'http';
});``

However, it may be beneficial to define a constant in wp-config.php so an additional plugin isn't necessary for non-pantheon development.

danielbachhuber:
> Defining environmental variables would be good to add as well:

Good suggestion — updated 288c5836028331863aa5637774375982b36ec80f

> However, it may be beneficial to define a constant in wp-config.php so an additional plugin isn't necessary for non-pantheon development.

Yeah, it might actually be nice to have all of these configuration details configurable by constants.

ataylorme:
> `add_filter('solr_scheme',function(){ return 'http'; });`

> However, it may be beneficial to define a constant in wp-config.php so an additional plugin isn't necessary for non-pantheon development.

@allan23 I like the new constant idea. Check out [PR #96](https://github.com/pantheon-systems/solr-power/pull/96)

allan23:
After reviewing my local setup, I also have another constant in wp-config.php:
``define( 'SOLR_PATH', '/solr/' );``
This sets the path to Solr on non-Pantheon environments. Per your instructions to use the install script, the developer should be okay, because it works in TravisCI.

README.md
```diff
@@ -24,12 +24,30 @@ Search is critical for your site, but the default search for WordPress leaves a
...
+If you're using the Solr Power plugin on Pantheon, setting up Apache Solr is as easy as enabling the Apache Solr add-on in your Pantheon dashboard. Once you've done so:
+
+1. Index your existing content by going to the plugin options screen and clicking "Execute" on "Index Searchable Post Types".
+2. Search on!
+3. See the examples/templates directories for more rich implementation guidelines.
+
+If you're using the Solr Power plugin elsewhere, you'll need to install and configure Apache Solr. On a Linux environment, this involves three steps:
+
+1. Install the Java Runtime Environment.
+2. Run `./bin/install-solr.sh` to install and run Apache Solr on port 8983.
```
ataylorme: @danielbachhuber I was able to get setup with a newer version of Solr locally by using [Homebrew](brew.sh) to install Solr (`brew install solr && brew services start solr`). I then just took ownership of the log files with ```sudo chown -R `whoami` /usr/local/Cellar/solr/``` and Solr is working on port `8983`.

I've opened [PR #112](https://github.com/pantheon-systems/solr-power/pull/112) to update the Solr version in `install-solr.sh` but using homebrew might be easier for some. What do you think about listing it as an option?

danielbachhuber: > What do you think about listing it homebrew as an option?

👍  Works for me. Feel free to edit as you see fit.

ataylorme: Homebrew only supports more recent version of Solr - I ended up reverting and installing Solr manually.

PR #84, opened 2016-07-15
Flesh out the Composer.json with more details

danielbachhuber:

PR #95, opened 2016-07-18
First pass at Behat integration tests for Solr Power

### LINKED ISSUE #79: Use Behat to test Solr Power on Pantheon
On a couple other Pantheon plugins now, we're using Behat to verify the plugin works as expected on Pantheon (e.g. https://github.com/pantheon-systems/wp-redis/pull/103). It would be cool if Solr Power did this too.

At a high-level, here's how it works:

* We're using Circle (in addition to Travis) to automate the creation and destruction of a Pantheon environment for an existing Pantheon site. In accordance to our existing naming scheme, you'd use `solr-power` as the site name.
* `bin/behat-prepare.sh` creates the environment, pushes the current plugin copy up, activates it, and performs any other configuration steps needed.
* `bin/behat-test.sh` runs the Behat tests against the created environment
* `bin/behat-cleanup.sh` removes the environment after the tests are done running.

You can likely copy and lightly modify most of the referenced pull request. I'd be keen to get your input on the documentation, and implementation generally.

cc @stevector

danielbachhuber:
See #79

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,28 @@
+machine:
+  php:
+    # https://circleci.com/docs/environment#php
+    version: 5.6.22
+  environment:
+    TERMINUS_ENV: ci-$CIRCLE_BUILD_NUM
+    TERMINUS_SITE: solr-power
+
+dependencies:
+  cache_directories:
+    - ~/.composer/cache
+  pre:
+    # Set the PHP timezone so that Behat script does not fail.
+    # Using > instead of >> will overwrite the file and disable xdebug.
+    # xdebug makes composer slower.
+    - echo "date.timezone = 'US/Central'"  >  /opt/circleci/php/5.6.22/etc/conf.d/xdebug.ini
+  override:
+    - composer global require pantheon-systems/terminus
+    - composer install
+    - terminus auth login --machine-token=$TERMINUS_TOKEN
+
+test:
+  pre:
+    - ./bin/behat-prepare.sh
+  override:
+    - ./bin/behat-test.sh
+  post:
+    - ./bin/behat-cleanup.sh
--- a/circle.yml
+++ b/circle.yml
@@ -16,12 +16,14 @@ dependencies:
     - echo "date.timezone = 'US/Central'"  >  /opt/circleci/php/5.6.22/etc/conf.d/xdebug.ini
   override:
     - composer global require pantheon-systems/terminus
-    - composer install
     - terminus auth login --machine-token=$TERMINUS_TOKEN
 
 test:
   pre:
     - ./bin/behat-prepare.sh
+    # Only run after the current directory has been deployed to Pantheon
+    # See https://github.com/pantheon-systems/solr-power/issues/72#issuecomment-234073642
+    - composer install
   override:
     - ./bin/behat-test.sh
   post:
--- a/behat.yml
+++ b/behat.yml
@@ -1,0 +1,12 @@
+# behat.yml
+default:
+  suites:
+    default:
+      paths:
+        - tests/behat
+      contexts:
+        - Behat\MinkExtension\Context\MinkContext
+  extensions:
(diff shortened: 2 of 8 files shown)


danielbachhuber:
#84 is going to conflict with this PR's dependencies, so I haven't added them yet.

danielbachhuber:
@stevector This is ready for a review. Note: I didn't commit the changes to the `/vendor/` directory. Not sure what we want to do there.

circle.yml
```diff
@@ -0,0 +1,28 @@
...
+    - composer install
+    - terminus auth login --machine-token=$TERMINUS_TOKEN
+
+test:
+  pre:
+    - ./bin/behat-prepare.sh
+  override:
+    - ./bin/behat-test.sh
+  post:
+    - ./bin/behat-cleanup.sh
```
stevector: @danielbachhuber Can you comment this out until tests are passing?

danielbachhuber: Here's the source of the failure:

image

What should we do about the vendor directory generally?

circle.yml
```diff
@@ -0,0 +1,28 @@
+machine:
+  php:
+    # https://circleci.com/docs/environment#php
+    version: 5.6.22
```
stevector: I found in some other repos that PHP 7 inside Circle and prestissimo make build slightly faster: https://github.com/pantheon-systems/search_api_pantheon/pull/25/files

circle.yml
stevector: From our meeting yesterday with @ataylorme we agreed to that the ideal set up for the vendor dir would be:

- No vendor directory in github.
- Treat wordpress.org's SVN as a repository of build artifacts that would include the full vendor directory.
- Some degree of automation sending build artifacts to SVN in response to tags on github.

However, right now our master branch here on github does have the vendor directory fully committed. I think for the scope of this PR we should not be changing anything related to the vendor directory unless we have to. The initial PR should simply automate (what I assume is) an already functional set of install instructions.

Having never installed this plugin myself, I'm not certain if this failure is because something in the vendor directory needs to change or something in the set up script needs to change.

circle.yml
danielbachhuber: > I found in some other repos that PHP 7 inside Circle and prestissimo make build slightly faster:

Huh. Because Composer is faster?

stevector: Yes, prestissimo makes composer faster by doing downloads in parallel. PHP7 should make all PHP-based scripts faster, Composer and Terminus.

circle.yml
danielbachhuber: > Treat wordpress.org's SVN as a repository of build artifacts that would include the full vendor directory.

How do we accommodate the fact that some dependencies are only for testing (e.g. development), and don't make sense in the final build artifact?

> Having never installed this plugin myself, I'm not certain if this failure is because something in the vendor directory needs to change or something in the set up script needs to change.

@ataylorme Mind taking a look at this failure and weighing in with your thoughts?

stevector: > How do we accommodate the fact that some dependencies are only for testing (e.g. development), and don't make sense in the final build artifact?

We use Composer's distinction between `require` and `require-dev`. When building for SVN we would use `composer install --no-dev`

danielbachhuber:
@stevector @ataylorme Fixed this up. Could I get a #reviewmerge so I can move on, and we can fix the failed builds on the other PRs?

ataylorme:
> @stevector @ataylorme Fixed this up. Could I get a #reviewmerge so I can move on, and we can fix the failed builds on the other PRs?

@danielbachhuber Thanks for this. The tests in `tests/behat/solr-power.feature` look good to me and I verify they passed on CircleCI.

Once we merge the facet widget we will want to expand tests to include indexing and searching for results but that's more of a would like to have.

The current checks that the plugin is loaded by visiting the options page and that Solr is available are good for now.

Issue #103, opened 2016-07-22
Hide UI Option to Submit Schema on non-Pantheon Environments


danielbachhuber:
The `SolrPower_Api->submit_schema()` method has some hardcoded dependencies on Pantheon, which makes it very difficult to use in other environments. We should decouple it from the Pantheon environment dependency.

allan23:
@danielbachhuber Can you elaborate on this issue? The submit_schema() method is only for Pantheon and is not an option on non-Pantheon environments. I don't foresee any other environment allowing the ability to send the schema.xml to the Solr instance.

When you say decouple it, are you saying to remove this method completely from the plugin? And if so, does that mean that Pantheon should have separate functionality only on their platform?

I mean it can be done (maybe add some hooks to the options page for Pantheon specific features), just need some clarification on this because it was a must-have feature for this project.

(CC: @joshkoenig  @ataylorme)

danielbachhuber:
> Can you elaborate on this issue? The submit_schema() method is only for Pantheon and is not an option on non-Pantheon environments. I don't foresee any other environment allowing the ability to send the schema.xml to the Solr instance.

Ok, I wasn't aware of this. Should we hide the UI in a non-Pantheon environment then?

> When you say decouple it, are you saying to remove this method completely from the plugin? And if so, does that mean that Pantheon should have separate functionality only on their platform?

Not necessarily. More so, I saw it as a hardcoded dependency on Pantheon that would benefit from some form of abstraction. Given the detail you've provided, I think it would be fine to just hide the UI in a non-Pantheon environment.

ataylorme:
>  just hide the UI in a non-Pantheon environment.
+1 for this - should be a simple change.

Issue #75, opened 2016-06-11
Add wp-cli support


ataylorme:
We should support actions from the plugins settings page in wp-cli.
* Check Server Settings
* Optimize Index
* Delete All
* Repost schema.xml
* Index Searchable Post Types

We should also consider informative commands.
* `info` should return ping status, Solr server IP address, Solr server port and Solr server path.
* `stats` should return a table of indexing status by post type

I think these two would create a minimum viable product. Additional commands can be items from the indexing tab in the plugin settings such as setting the default search operator.

danielbachhuber:
@ataylorme Want some help with this?

ataylorme:
As @allan23 pointed out there is some initial WP-CLI support ([see here](https://github.com/pantheon-systems/solr-power/blob/master/includes/class-solrpower-cli.php)) but the coverage is lacking.

We would like to expand the coverage as well as add a wiki page to Github with documentation on available commands.

ataylorme:
Merged in #104

ataylorme:
Documentation added to `README.md`. Since this is displayed on the repo home page I don't think there is a need to add to the wiki as well.

PR #104, opened 2016-07-22
Introduce more WP-CLI commands for Solr Power

danielbachhuber:
* Refactors `wp solr delete` to accept one or more post ids, or the `--all` flag. Previously, not passing any arguments would cause the index to be deleted, which is an implicit behavior. It's better to have an explicit interface.
* Introduces `wp solr optimize-index` for optimizing the index.
* Introduces `wp solr check-server-settings` which pings the server to see if the connection is valid.
* Adds `wp solr info` and `wp solr stats`, which display tables about server info and index stats, respectively.
* Introduces `wp solr repost-schema` for reposting the schema.xml to the server. Unfortunately, this only works on Pantheon right now (#103)

See #75

### DIFF
--- a/includes/class-solrpower-cli.php
+++ b/includes/class-solrpower-cli.php
@@ -1,6 +1,72 @@
 <?php
 
+/**
+ * Perform a variety of actions against your Solr instance.
+ */
 class SolrPower_CLI extends WP_CLI_Command {
+
+	/**
+	 * Check server settings.
+	 *
+	 * Pings the Solr server to see if the connection is functional.
+	 *
+	 * @subcommand check-server-settings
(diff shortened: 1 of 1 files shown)


pantheon-circleci:
@danielbachhuber, thanks for your PR! By analyzing the blame information on this pull request, we identified @allan23 to be a potential reviewer

danielbachhuber:
@ataylorme #reviewmerge?

ataylorme:
> Previously, not passing any arguments would cause the index to be deleted, which is an implicit behavior. It's better to have an explicit interface.

Agreed - thanks for the update.

PR #105, opened 2016-07-24
Solarium cleanup

ataylorme:
Defork Solarium (closes #72) by creating the `PantheonCurl` class and update Solarium to 3.6.0 (closes #14).

Issue #72, opened 2016-04-27
De-fork Solarium


joshkoenig:
Pre-requisite for #14 - we have a little shim in Solarium. We should subclass instead like the proper OOPers that we are.

ataylorme:
Removing the 0.4 milestone. Since this isn't a new feature, just some internal cleanup, it can be a patch version bump instead of a minor release and shouldn't hold up other, new features for 0.4

ataylorme:
@allan23 Check out @stevector's [commit on removing the Solarium patch for the Drupal module](https://github.com/stevector/search_api_pantheon/pull/19/files#diff-ab66d9a69b5ed6f6025041ee470895d5R1).

stevector:
I did some cleanup on the subclass: https://github.com/stevector/search_api_pantheon/pull/21/files

ataylorme:
Thanks Steve!

danielbachhuber:
@ataylorme @allan23 Does this mean it's not safe to run `composer install`, because there's some forked code committed to the `vendor` directory?

ataylorme:
@danielbachhuber Started this in PR #105. Would love to chat with you or @greg-1-anderson on getting our `vendor` directory out of the repo.

ataylorme:
Merged in #106

Issue #15, opened 2015-07-14
Script to go from GitHub repo to WordPress SVN


joshkoenig:
We need a script for converting this github project to publish via SVN to wp.org/plugins. There are a few already out there, eg:

http://scribu.net/blog/deploying-from-git-to-svn.html

https://danielbachhuber.com/2012/09/30/git-in-my-subversion/

And some tools that could help:

https://github.com/benbalter/WP-Readme-to-Github-Markdown

It would be great to make something that we can share w/others on this. Would be an excellent best-practice to share. 

The goal is to have a "scripts" directory in the plugin that includes a shell script to do the translation/release in a repeatable way. We can rely on the local script-running user having all appropriate access credentials. 

**Bonus Points** it would be great to lay out a pattern for doing development work on Pantheon as well, pulling changes from a Pantheon site repo into this repo via `git subtree merge`. That would be a separate item in the scripts directory.

joshkoenig:
More prior art to take into account (thx to @afragen for these!)

https://github.com/dd32/Github-to-WordPress-Plugins-Sync

https://github.com/GaryJones/wordpress-plugin-git-flow-svn-deploy

danielbachhuber:
> /me puts on his curmudgeon hat

For a long time now, I've wanted an automated way to deploy to WordPress.org SVN from Github. `master` branch would automatically sync with trunk, and tagged releases would magically appear as SVN tags. And yet, I still do it manually through the approach documented in my blog post.

As a plugin maintainer, my chief responsibility is to ensure every release is as bug-free as I can make it. Because I haven't come across, or produced, a Git to SVN script I trust to work reliably, I consider it my responsibility to manually produce and verify each release. Furthermore, I've run into race condition problems on Travis which has meant an old, slower build deploys over a newer, already completed build.

The REST API also has a release script which I believe Ryan runs manually https://github.com/WP-API/WP-API/blob/develop/bin/release.sh We could package something similar, although I think it's critically important the maintainer does a final review of the code to be committed.

> It would be great to make something that we can share w/others on this. Would be an excellent best-practice to share.

Conceivably, we could put this in WP-CLI too.

joshkoenig:
I don't disagree that testing and review should be a part of the release process. What I'm looking for is a standard process for doing this so that it's done consistently, and that the repeatable parts of it can be automated.

[...]

Issue #14, opened 2015-07-14
Update Solarium to 3.5.1


joshkoenig:
We should get up to 3.3.0. However this is not a priority for the first (alpha) release.

joshkoenig:
It's up to 3.5.1 now! We should also standardize a build/test process here. 

https://github.com/solariumphp/solarium/releases

joshkoenig:
Also, make sure we've removed any hacks that went into Solarium.

ataylorme:
Removing the 0.4 milestone. Since this isn't a new feature, just some internal cleanup, it can be a patch version bump instead of a minor release and shouldn't hold up other, new features for 0.4

ataylorme:
Merged in #106

PR #106, opened 2016-07-24
Solarium cleanup

ataylorme:
Continuation of #105


greg-1-anderson:
You can avoid double test runs on Travis like so:
```
branches:
  only:
    - master
    - /^[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+.*$/
```
Without this, the `push` rule will match for pull requests, since the pull request is a branch also.  n.b. The regex at the end matches branches that are SemVer versions. This causes release tags to be tested rather than ignored (since the `branches` directive applies to tags in addition to branches).

ataylorme:
> Why did you switch from composer to curl for installing Terminus? The former has the advantage of being able to select the most recent version, whereas the curl replacement selects a specific version of terminus, so you won't necessarily get bugfixes over time.

CircleCI was failing with the message "bash: line 1: terminus: command not found" when installing with Composer.

Maybe because the Composer bin directory isn't added to PATH?

circle.yml
```diff
@@ -16,6 +16,7 @@ dependencies:
     - echo "date.timezone = 'US/Central'"  >  /opt/circleci/php/5.6.22/etc/conf.d/xdebug.ini
   override:
     - composer global require pantheon-systems/terminus
+    - echo 'export PATH=~/.composer/vendor/bin:$PATH' >> ~/.bashrc
```
greg-1-anderson: .bashrc has already been sourced, and won't be sourced again unless you explicitly do so.  So, either add another line, `source ~/.bashrc`, or just change the line above to something like:

```
PATH=PATH=~/.composer/vendor/bin:$PATH
```

ataylorme:
This addresses #72 and #14 by de-forking Solarium and updating to the latest version.

@greg-1-anderson I added the `vendor` directory back to make the PR for readability. Once we merge this I'll open a second pull request to ditch the `vendor` directory. Currently Composer is installing over the top of the old `vendor` directory.

cc @danielbachhuber

ataylorme:
@allan23 Please review when you get a chance

bin/behat-prepare.sh
```diff
@@ -45,7 +45,9 @@ rm -rf $PREPARE_DIR/wp-content/plugins/solr-power/.git
 # Push files to the environment
 ###
 cd $PREPARE_DIR
-git add wp-content
+git rm -rf --cached wp-content/plugins/solr-power/vendor
+find $PREPARE_DIR/wp-content/plugins/solr-power/vendor -type d -name '.git' | xargs rm -rf
```
ataylorme: When pushing to the Pantheon multidev it throws an error message about submodules due to having `.git` directories inside the `vendor` directory, even though they aren't submodules, there is no `.gitmodules` file, so the offending `.git` directories must be removed.

bin/behat-prepare.sh
```diff
@@ -45,7 +45,9 @@ rm -rf $PREPARE_DIR/wp-content/plugins/solr-power/.git
 # Push files to the environment
 ###
 cd $PREPARE_DIR
-git add wp-content
+git rm -rf --cached wp-content/plugins/solr-power/vendor
```
ataylorme: Since we have an old `vendor` directory committed it needs to be purged from Git cache.

circle.yml
```diff
@@ -16,14 +16,13 @@ dependencies:
     - echo "date.timezone = 'US/Central'"  >  /opt/circleci/php/5.6.22/etc/conf.d/xdebug.ini
   override:
     - composer global require pantheon-systems/terminus
+    - PATH=PATH=~/.composer/vendor/bin:$PATH
```
ataylorme: Terminus wasn't being recognized as a command - this is the fix.

circle.yml
```diff
@@ -16,14 +16,13 @@ dependencies:
...
     - terminus auth login --machine-token=$TERMINUS_TOKEN
 
 test:
   pre:
-    - ./bin/behat-prepare.sh
-    # Only run after the current directory has been deployed to Pantheon
-    # See https://github.com/pantheon-systems/solr-power/issues/72#issuecomment-234073642
     - composer install
+    - ./bin/behat-prepare.sh
```
ataylorme: We want to run `composer install` before deploying to Pantheon so we test against the updated version of Solarium

solr-power.php
```diff
@@ -126,13 +112,16 @@ function solr_power__deactivate() {
 	add_action( 'admin_notices', 'solr_power_env_variables_admin_notice' );
 	add_action( 'admin_init', 'solr_power__deactivate' );
 } else {
-	define( 'SOLR_POWER_PATH', plugin_dir_path( __FILE__ ) . '/' );
+	define( 'SOLR_POWER_PATH', untrailingslashit( plugin_dir_path( __FILE__ ) ) );
```
ataylorme: There is a slash in the require statements below, having it here was redundant and caused duplicates.

circle.yml
greg-1-anderson: Typo here: `PATH=PATH=`

solr-power.php
stevector: Instead of manually requiring all of these files, how about using Composer autoloading?

ataylorme: @stevector That's a good idea - I added this as a [issue #110](https://github.com/pantheon-systems/solr-power/issues/110)

ataylorme:
@allan23 Greg helped me clean up an issue with `.git` directories in the `vendor` directory of the CirlceCI build. All set now and this is ready for review.

composer.json
```diff
@@ -14,7 +14,7 @@
     },
     "minimum-stability": "dev",
     "require": {
-        "solarium/solarium": "3.2.0"
+        "solarium/solarium": "3.6.0"
```
allan23: Since this is updating to Solarium 3.6.0 via composer, shouldn't Solarium be updated in the repo or should the vendor directory be removed?

If the latter, gitignore needs to be updated to ignore vendor.

allan23:
@ataylorme all appears to be working. My only concern was the Solarium version and vendor directory.

composer.json
ataylorme: > If the latter, gitignore needs to be updated to ignore vendor.
I'm in favor of removing the `vendor` directory and running `composer install` when running tests, as this PR does, and before deploying to WordPress.org since it will need to exist there.

@joshkoenig any objection to dropping the `vendor` directory from the Git repo?

stevector: The sooner the better for removing `vendor` IMO. Right now there is a mismatch in that `vendor` is version controlled but `composer.lock` is not. I think there is agreement that eventually neither should be version controlled here and that both would be added by a packaging script for wordpress.org: https://github.com/pantheon-systems/solr-power/issues/15

I think there should be a separate PR that removes the vendor dir and updates the README. I say a separate PR because it will have a huge diff that would make other file changes difficult to review.

ataylorme: >  think there should be a separate PR that removes the vendor dir and updates the README.

Yes, I agree it should be a separate PR. I'll open one later today.

ataylorme: Added [PR #117](https://github.com/pantheon-systems/solr-power/pull/117)

PR #109, opened 2016-08-03
Using test and cleanup scripts from pantheon-wordpress-upstream-tests

stevector:
@danielbachhuber, when we started looking at Behat tests for the WordPress plugins we wondered how we could reduce code duplication. What do you think of reusing the test and cleanup scripts? cc @ataylorme @greg-1-anderson


pantheon-circleci:
@stevector, thanks for your PR! By analyzing the blame information on this pull request, we identified @danielbachhuber to be a potential reviewer

greg-1-anderson:
Overall seems like a good idea to me. I do wonder if there's anything we need to do to keep track of these cross dependencies, as re-organizing the pantheon-wordpress-upstream-tests would break everything downstream that used this pattern. A while back I made some testing projects, ci-scripts, travis-scripts, circle-scripts, etc., and the thing that I did there was declare all of the test scripts to be Composer binaries, so they all ended up in the bin dir (defaults to `vendor/bin`, but I relocated it to just `bin`, because I had relocated `vendor` to a deep folder for Drupal 7 w/ no docroot relocation). Test scripts mixed in with other scripts in your bin directory might be a little messy, but on the flip side, most sites don't put much in their bin dir.

Whatever we do, we should make a plan on how to document and track these relationships. Otherwise, the duplication would be preferable to the maintenance cost.

stevector:
> so they all ended up in the bin dir

Yes, I think that's preferable and would reduce the fragility of the coupling.

> Whatever we do, we should make a plan on how to document and track these relationships. Otherwise, the duplication would be preferable to the maintenance cost.

Yes, DRY is a form of coupling and if the coupling isn't understood/documented it can be worse than duplicate code. Most of my practical and theoretical knowledge in this problem space is in PHP itself rather than Bash scripts. From that world I know part of the answer can be semantic versioning and interfaces. Introducing semantic version numbers to `pantheon-wordpress-upstream-tests` could help. Defining the interfaces of Bash scripts, I don't know how to do that.

danielbachhuber:
> What do you think of reusing the test and cleanup scripts?

I suppose this is fine. I don't see a ton of benefit to abstraction in this particular case, but we could establish the pattern of using the scripts from the upstream by default, and then internalizing if needed.

> Defining the interfaces of Bash scripts, I don't know how to do that.

We may be in uncharted territory :)

stevector:
@danielbachhuber let's not merge this PR

stevector:
This repo is the last of four plugins repos to get this integration. None of the first three dropped their existing test/cleanup scripts. There's no rush to remove these files. I'd rather think on it a little while longer.

danielbachhuber:
> There's no rush to remove these files. I'd rather think on it a little while longer.

That's fine by me. Want to open an issue in the upstream repo about general abstraction?

stevector:
Yes, here's a place holder: https://github.com/pantheon-systems/pantheon-wordpress-upstream-tests/issues/29

PR #112, opened 2016-08-04
Update Solr version to `5.5.2`

ataylorme:
Tests should use the same version of Solr as Pantheon. Solr `5.3` and above don't use a `schema.xml` anymore.

### DIFF
--- a/bin/install-solr.sh
+++ b/bin/install-solr.sh
@@ -42,14 +42,11 @@ post_some_documents() {
 download_and_run() {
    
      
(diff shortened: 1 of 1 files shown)


allan23:
@ataylorme When did Pantheon update?  

I was told that Pantheon did not have plans to update from 3.6 to 5.x any time soon.

ataylorme:
@allan23 I misread `3.5` as `5.5` - too early I guess. We are indeed on `3.6`. Going to close this out.

PR #117, opened 2016-08-05
Remove vendor directory

ataylorme:
As part of Solarium Cleanup ([PR #106](https://github.com/pantheon-systems/solr-power/pull/106)) the vendor directory is no longer needed.

`composer install` should be run on CI before testing and manually before deploying to WordPress.org.

Also adding `composer.lock` to the Git repository.


ataylorme:
Approved by Josh, merging this in.

PR #118, opened 2016-08-05
Facet widget clean

ataylorme:


ataylorme:
@allan23 Creating a new PR for #78 with cleaner history. That one got mucked up when rebasing `master` which deleted the `vendor` directory.

ataylorme:
Testing this  when I look at the product category "Movie & TV" with the search "apple" there are 5 results: Fish Violet Gift, Lime Jazz Cold, Green Hair Cover, ACME Brass, and Seat Cleaner Samsung. Of these 5 results only Fish Violet Gift has the tag fish. 

With the AND operator I would expect faceting the "apple" search for the product category "Movie & TV" and the tag "Fish" to return only the Fish Violet Gift result as it is the only item from the "apple" search result set that has the "Movie & TV" category and the tag "Fish".

Instead I am seeing a result set that includes the results for "apple" that have the fish tag combined with the result set that has the "Movie & TV" category.

When the AND operator is set we want to use AND for the different groups of facets but keep OR for facets of the same type. 

For example, searching for "apple" and faceting on the product categories "Movie & TV" as well as "Board Games" and the tag "ex" should display all results for the search term "apple" that have either: A) The "Movie & TV" category AND the tag "ex" or B) The "Board Games" category AND the tag "ex"

Going back to the car example we want user to be able to search multiple facets of the same type but still limit by many facets. e.g. Cars with the model years 2014 or 2015 AND that are red or blue.

ataylorme:
@allan23 This is all looking good to me. Anything else you can think of before I merge?

allan23:
@ataylorme Not that I can think of.

Issue #83, opened 2016-07-14
Documentation Needed For Filters


allan23:
It will be beneficial for developers to have all filters in this plugin documented on the GitHub Wiki.

ataylorme:
Closed this prematurely. We still need to document the filters on the project Wiki.

Ritesh-patel:
Added wiki page https://github.com/pantheon-systems/solr-power/wiki

ataylorme:
Great, thank you @Ritesh-patel. Can you also add this to `readme.txt` and `README.md`? Note that if you modify `readme.txt` there is a Grunt task to generate `README.md`

PR #119, opened 2016-08-09
Add incline documentation to existing hooks.

ChrisWiegman:
Adds inline documentation to all 8 existing filters used in the plugin.

Before I add these to the wiki per #83 here is a PR to add them to the code itself. Once confirmed that this is all accurate I will expend to a Wiki page.


ataylorme:
@ChrisWiegman there are also some new filters in the facet widget PR, such as [solr_index_custom_fields](https://github.com/pantheon-systems/solr-power/blob/a57a03173b1c9382fe050788a4fc581d2b6eac0e/includes/class-solrpower-sync.php#L183) that need documented as well.

I think the best course would be for @allan23 to update that PR to add inline documentation for new filters not covered here. Once both of these PRs are merged we'll have full coverage and can make a Wiki page.

Thoughts?

ChrisWiegman:
@ataylorme sounds good to me. If you want to merge this for what it has that's fine. If not, let's have @allan23 merge first and I'll add the remaining filter doc to this PR. Either direction is fine with me.

ataylorme:
I added inline docs for the new filters in commit [c9473462394a85aa84dd623be252b6007529ab67](https://github.com/pantheon-systems/solr-power/pull/118/commits/c9473462394a85aa84dd623be252b6007529ab67) of #118.

@allan23 assigning you here to review these filter docs as well as those in that commit.

If all looks good here we can merge this PR.

Issue #98, opened 2016-07-19
Abstract CSS and JS to dedicated files


danielbachhuber:
There is some inline CSS and JS in the options page that would be better abstracted out to dedicated resource files.

PR #120, opened 2016-08-09
Feature/options styles

ChrisWiegman:
Addresses #98 

Moves all inline JS and CSS from the options page to external resources and applies processing to each for better efficiency and easier debugging going forward.


ataylorme:
@ChrisWiegman I don't think we should commit compiled assets, only source files.

Can you: 
- remove the compiled assets
- add the relevant Grunt tasks to Travis and Circle CI
- update the README to include instructions for compiling the production assets with Grunt. it would be good to include how to use the watch command under the local installation section as well.

ataylorme:
Hey @ChrisWiegman,
> As for the CI runners, I'm not sure what the use case is here? We're not running any JS tests nor will the absence of any of the compiled tasks cause a unit test to fail. Is there something I'm missing there?

More of a sanity check here since Grunt failing will result in the tests failing - in which case we can go fix Grunt.

I'd also like the keep the CI runner scripts up to date for building the entire plugin so they can be a reference point when we get to #15.

That being said it's not crucial we add the Grunt task to the CI builds.

ChrisWiegman:
@ataylorme Added build of the default grunt task to both travis and circle

ataylorme:
> Added build of the default grunt task to both travis and circle

@ChrisWiegman Great, thanks. I added `node_modules` to cache for both CircleCI and Travis. Going to merge this in.

Issue #102, opened 2016-07-22
Create a public interface for getting server info


danielbachhuber:
Rather than having the following throughout:

```
$server = array(
	'ping_status' => $ping ? 'successful' : 'failed',
	'ip_address'  => getenv( 'PANTHEON_INDEX_HOST' ),
	'port'        => getenv( 'PANTHEON_INDEX_PORT' ),
	'path'        => SolrPower_Api::get_instance()->compute_path(),
);
```

We should have a method on the API class that returns this data, for use by the options page and CLI.

PR #123, opened 2016-08-10
Feature/api status

ChrisWiegman:
This is to address #102. Note that there is a lot of room for interpretation on the original issue so let me know if this is the direction you had in mind and how far that issue itself should go.

### DIFF
--- a/includes/class-solrpower-api.php
+++ b/includes/class-solrpower-api.php
@@ -461,4 +461,24 @@ class SolrPower_Api {
 
 		}
 	}
+
+	/**
+	 * Returns an array of server information.
+	 *
+	 * @since 1.0
+	 *
+	 * @return array Array of server connection information.
+	 */
+	public function get_server() {
+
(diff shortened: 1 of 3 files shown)


ataylorme:
@danielbachhuber can you review?

includes/class-solrpower-api.php
```diff
@@ -461,4 +461,24 @@ function check_for_schema() {
...
 	}
+
+	/**
+	 * Returns an array of server information.
+	 *
+	 * @since 1.0
+	 *
+	 * @return array Array of server connection information.
+	 */
+	public function get_server() {
```
danielbachhuber: Could we call this `get_server_info()`? `get_server()` is somewhat ambiguous — to me, it implies it will return a server object.

ataylorme: +1 for the clarity

ChrisWiegman:
@danielbachhuber the method name has been updated.

Issue #85, opened 2016-07-15
Use WPCS to ensure the codebase adheres to WordPress Coding Standards


danielbachhuber:
WPCS (https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) is a series of PHP Codesniffer Sniffs to ensure a codebase adheres to WordPress' coding standards. We should set it up for the project, as we already have for https://github.com/pantheon-systems/wp-redis et al.

Setting up WPCS involves:
- [x] Defining the dependencies in composer.json https://github.com/pantheon-systems/wp-redis/blob/master/composer.json#L11-L12
- [x] Adding the `grunt phpcs` step, and defining associated dependencies https://github.com/pantheon-systems/wp-redis/blob/master/Gruntfile.js#L20-L28
- [x] Having Travis call the Codesniffer step https://github.com/pantheon-systems/wp-redis/blob/master/.travis.yml#L39-L43

Once it's set up, we'll need to:

* [x] Update all files to adhere to WPCS

ataylorme:
I agree wo should be using [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/).

Outside of automation and formatting updates should be made independently of other commits.

ChrisWiegman:
I'd be happy to run with this as it will stop my IDE from yelling at me every time I save a file. Feel free to have @allan23 assign to me.

ataylorme:
@ChrisWiegman I assigned most items to @allan23 but feel free to grab anything from the [1.0 milestone](https://github.com/pantheon-systems/solr-power/milestone/5) to work on.

ataylorme:
Punting this so we can get `1.1` out.

ataylorme:
When working on this the first priority should be code comments. Once all code is properly commented/detailed we can work on formatting.

danielbachhuber:
🚢  :shipit: 🎉

PR #127, opened 2016-08-11
Feature/wordpress coding standards

ChrisWiegman:
This pull request addresses #85 

In lieu of adding to CI I propose we do two things:
1. Clean up the extraordinary amount of errors to avoid false positives due to so many existing errors being present.
2. Utilize @stevegrunwell's WP_Enforcer library (https://github.com/stevegrunwell/wp-enforcer). This moves the linting from CI to commit and will simply prevent a dev from committing code not up to standards thereby providing earlier feedback to the dev and preventing CI from throwing a failure for a syntax or similar error.


ataylorme:
@ChrisWiegman 

> 1) Clean up the extraordinary amount of errors to avoid false positives due to so many existing errors being present.

Yes, the existing errors will need to be cleaned up.

> 2) Utilize @stevegrunwell's WP_Enforcer library (https://github.com/stevegrunwell/wp-enforcer). This moves the linting from CI to commit and will simply prevent a dev from committing code not up to standards thereby providing earlier feedback to the dev and preventing CI from throwing a failure for a syntax or similar error.

This makes sense to me. After We clean up all existing code formatting issues disallowing new issues to be committed will prevent the codebase from having this many code standard issues in the future.

We'll continue down this path but this issue is low priority and should only be attended to after other, higher priority issues are complete.

danielbachhuber:
> In lieu of adding to CI I propose we do two things:

Is there any harm in running PHP Codesniffer on CI in addition to the pre-commit hook?

stevector:
I'v never used it but PHPCS has the ability to auto fix violations: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically

jcage88:
Thanks for the thread here! I am creating my Wordpress website, and it will surely be very helpful for me. It is very cool indeed

allan23:
@ataylorme: this commit consists of cleaned up code that adheres to WordPress Coding Standards, phpcs for grunt (runs on TravisCI), WP Enforcer (phpcs pre-commit hook), and the addition of a bulk sync/indexing option that has sped up unit tests by 61% (instead of relying on the post_status change which adds additional time when there are more than one posts being created).

danielbachhuber:
Restarting this in #251

PR #130, opened 2016-08-25
Use a better unauthenticated check

danielbachhuber:
This way, the script can be run when `terminus` is already
authenticated, but `TERMINUS_TOKEN` isn't present

### DIFF
--- a/bin/behat-cleanup.sh
+++ b/bin/behat-cleanup.sh
@@ -4,12 +4,18 @@
 # Delete the Pantheon site environment after the Behat test suite has run.
 ###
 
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@ataylorme #reviewmerge

ataylorme:
> This way, the script can be run when terminus is already
authenticated, but TERMINUS_TOKEN isn't present

@danielbachhuber Assuming this is for local testing?

danielbachhuber:
> Assuming this is for local testing?

Correct, although this is a more precise check in any case.

PR #137, opened 2016-10-03
Fix single quote in facet issue

### LINKED ISSUE #125: Accented characters in facet names cause issues
Similarly to how we are escaping characters like ampersand we need to escape accented characters.

Facet names with accented characters, e.g. "Room Décor", return no results when there are some.

Ritesh-patel:
Fixes #125

### DIFF
--- a/bin/install-solr.sh
+++ b/bin/install-solr.sh
@@ -4,7 +4,7 @@ SOLR_PORT=${SOLR_PORT:-8983}
 
 download() {
     echo "Downloading solr from $1..."
-    curl -s $1 | tar x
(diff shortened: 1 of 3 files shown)


ataylorme:
@Ritesh-patel this is working well on [the Pantheon test site](http://facet-widg-solr-power.pantheonsite.io/?s=apple&facet%5Bproduct_cat_taxonomy%5D%5B%5D=All+Kids%27+Furniture+%26+Room+D%C3%A9cor).

Travis is failing though so I want to get that fixed before merging.

PR #140, opened 2016-10-07
Facets in full test search results

### LINKED ISSUE #126: Return facets in full text search results
Solr search should return matches for facets in full text search. For example, if a post has the category "Movie" it should return in the results for the query "movie".

Facets values, such as category, should be considered but scored less than post title or content.

Ritesh-patel:
Fixes #126

### DIFF
--- a/includes/class-solrpower-api.php
+++ b/includes/class-solrpower-api.php
@@ -261,50 +261,11 @@ class SolrPower_Api {
 
 
 		$response         = null;
-		$facet_fields     = array();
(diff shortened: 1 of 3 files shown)


allan23:
This PR may no longer be necessary if #158 works as expected.

ataylorme:
@allan23 I'm closing this out since it's half finished work and, as you said, #158 should take care of it. Feel free to re-open if you think there is more to be done here.

PR #141, opened 2016-10-07
Update readme: add wiki page link for documenration under Development

Ritesh-patel:


Ritesh-patel:
Updated readme for #83

PR #158, opened 2016-11-18
Allow post_title and post_content to score higher

### LINKED ISSUE #126: Return facets in full text search results

allan23:
Utilizing a boost function, post_title and post_content can now score higher in results instead of facet items such as taxonomies, custom fields, or categories. This would resolve #126.

### DIFF
--- a/tests/phpunit/test-solr.php
+++ b/tests/phpunit/test-solr.php
@@ -444,12 +444,26 @@ class SolrTest extends SolrTestBase {
 		wp_set_object_terms( $p_id, $cat_id2, 'category', true );
 
 
-		$p_id2 = $this->__create_test_post( 'post', 'Best Movies of 2015' );
+		$p_id2 = $this->__create_test_post( 'post', 'Best Films of 2015' );
 		wp_set_object_terms( $p_id2, $cat_id, 'category', true );
 
 		SolrPower_Sync::get_instance()->load_all_posts( 0, 'post', 100, false );
 		$args  = array(
 			's' => 'Movie'
+		);
+		$query = new WP_Query( $args );
+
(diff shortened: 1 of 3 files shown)


tests/phpunit/test-solr.php
```diff
@@ -401,4 +401,59 @@ function test_multi_facet_and() {
...
+	 * @group 126
+	 * @link https://github.com/pantheon-systems/solr-power/issues/126
+	 */
+	function test_facets_in_results_content() {
+		$cat_name  = 'Movies';
+		$cat_id    = wp_create_category( $cat_name );
+		$cat_name2 = 'Review';
+		$cat_id2   = wp_create_category( $cat_name2 );
+
+		$p_id = $this->__create_test_post( 'post', 'Casablanca Movie', 'Rated as one of the best movies of all time. This movie is epic.' );
```
ataylorme: @allan23 - since the post title here has _Movie_ in it we aren't strictly testing the category.

tests/phpunit/test-solr.php
```diff
@@ -401,4 +401,59 @@ function test_multi_facet_and() {
...
+
+		wp_set_object_terms( $p_id, $cat_id2, 'category', true );
+
+
+		$p_id2 = $this->__create_test_post( 'post', 'Best Movies of 2015' );
```
ataylorme: Ditto here - maybe change this to `Best Films of 2015` *or* search for the term `Review` below so we are testing a query against the faceted item only.

allan23:
@ataylorme - let me know if the latest additions to the unit tests are sufficient or if there should be more added.

ataylorme:
> let me know if the latest additions to the unit tests are sufficient or if there should be more added.
These look good to me

PR #159, opened 2016-11-29
date_query

ataylorme:
Adds `date_query` support to `WP_Query`

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -938,7 +938,7 @@ class SolrPower_WP_Query {
 					return '(' . $column . ':' . '[' . $the_date . 'T00:00:00Z TO *])';
 					break;
 				default:
-					return '(' . $column . ':' . $the_date . 'T00:00:00Z)';
+					return '(' . $column . ':' . $the_date . '*)';
 					break;
 			}
 
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -32,6 +32,8 @@ class SolrPower_WP_Query {
 	 */
 	var $fq = array();
 
+	var $query;
+
 	/**
 	 * Grab instance of object.
 	 * @return SolrPower_WP_Query
@@ -398,7 +400,8 @@ class SolrPower_WP_Query {
 	 * @return string
 	 */
 	private function build_query( $query ) {
-		$whitelist = array(
+		$this->query = $query;
+		$whitelist   = array(
 			'post_type',
 			'p',
 			'page_id',
@@ -407,7 +410,7 @@ class SolrPower_WP_Query {
 			'name'
 
 		);
-		$convert   = array(
+		$convert     = array(
 			'p'       => 'ID',
 			'page_id' => 'ID',
 			'name'    => 'post_name'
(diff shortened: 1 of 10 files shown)


ataylorme:
@allan23 Travis is failing when installing Solr. I think the download from `http://archive.apache.org/dist/lucene/solr/3.6.2/apache-solr-3.6.2.tgz` is getting corrupted. I've tried to re-run and few times and also switched download methods to no avail.

Have you seen this before? Any ideas?

cc @danielbachhuber

allan23:
@ataylorme That is definitely a new one. I wonder if TravisCI is getting flagged for too many requests?

Should we consider hosting the solr package instead of hitting up apache.org?

I'm curious to know that since the download is so small, it may be sending plaintext to Travis.

includes/class-solrpower-wp-query.php
```diff
@@ -822,6 +836,235 @@ private function set_query_value( $value, $type = null ) {
...
+			) {
+				$query[] = $this->parse_date_query( $dq );
+				continue;
+			}
+			$inclusive = ( isset( $dq['inclusive'] ) ) ? $dq['inclusive'] : false;
+
+			if ( isset( $dq['before'] ) && is_array( $dq['before'] ) ) {
+				$query[] = $this->date_query( $dq['before'], $inclusive, 'before' );
+			} elseif ( isset( $dq['before'] ) ) {
+				$the_date = $this->query->date_query->build_mysql_datetime( $dq['before'], $inclusive );
```
ataylorme: @allan23 when a string is passed to [`build_mysql_datetime`](https://developer.wordpress.org/reference/classes/wp_date_query/build_mysql_datetime/) it is run through `strtotime`. When using before/after we need to check for arrays (year/month/day) and pass an array to `build_mysql_datetime`.

Date queries such as:
```
[date_query] => Array
        (
            [0] => Array
                (
                    [before] => Array
                        (
                            [year] => 2000
                            [month] => 
                            [day] => 
                        )

                )

            [1] => Array
                (
                    [after] => Array
                        (
                            [year] => 1990
                            [month] => 
                            [day] => 
                        )

                )

            [relation] => AND
        )
```

Are querying solr for `(post_date:[* TO 1970-01-01T00:00:00Z])AND(post_date:[1970-01-01T00:00:00Z TO *])` because of this.

When day or month are omitted from a `date_query` array we need to use defaults and pass them to `build_mysql_datetime` in array, not string, format. The defaults are documented on the [`WP_Query` codex article](https://codex.wordpress.org/Class_Reference/WP_Query#Date_Parameters)

ataylorme: @allan23 the test passes on Travis. I think the error might be due to month and day being passed empty in the query arguments. When you switch to using an array of year/month/day for `build_mysql_datetime` rather than a string please check for empty values and revert to defaults.

allan23: @ataylorme your unit test is passing with the Solr query being ```((year_i:[* TO 1990])AND(year_i:[1987 TO *]))``` 

WP core throws ```_doing_it_wrong()``` notices when date values are not correct (empty strings, 13 as month, etc.).

Is there something I'm missing in order to get the same Solr query as above?

ataylorme: @allan23 

> WP core throws _doing_it_wrong() notices when date values are not valid (empty strings, 13 as month, etc.).

Good to know

> Is there something I'm missing in order to get the same Solr query as above?

Nope, I think we're all set here. This was just a matter of the template created to test passing empty values for day/month.

PR #160, opened 2016-11-29
Use wget rather than curl to download Solr

ataylorme:


ataylorme:
> @ataylorme That is definitely a new one. I wonder if TravisCI is getting flagged for too many requests?

@allan23 I think you are right about rate limiting. I switched to `wget` and am showing a 503. See [this build](https://travis-ci.org/pantheon-systems/solr-power/jobs/179902212).

> Should we consider hosting the solr package instead of hitting up apache.org?

I think hosting it may be a bit much. Instead I'm going to lay off for a while then try again but use Travis' caching capabilities to cache the downloaded directory in the CI environment.

ataylorme:
There are [other mirrors available](https://www.apache.org/dyn/mirrors/mirrors.cgi) but they only host recent versions, not `3.6.2`

Issue #124, opened 2016-08-10
Ajax facet search support


ataylorme:
User's expect a modern experience and having to scroll back to the top of the facet widget and click search to see updated results is a poor experience.

We should add some Ajax to the facet widget so when facet select inputs are changed after the initial search the results are updated.

The facet inputs should also be disabled while new results are being fetched so another ajax call isn't fired off before the first one finished.

ataylorme:
> when a facet is selected, the search results should update. However, the results are based upon the search template in the theme and not everyone will be using the same theme. One solution I can think of would be to create a default search result layout in the plugin that can be overwritten in the theme. What are your thoughts on this?

@allan23 adding your comments from email here. I'll need to think about this and get back to you.

ataylorme:
@allan23,
> One solution I can think of would be to create a default search result layout in the plugin that can be overwritten in the theme. What are your thoughts on this?

I like this approach. We can use the [template_include filter](https://codex.wordpress.org/Plugin_API/Filter_Reference/template_include) and provide a default search template with the plugin but check if an override exists in the theme directory. Something like "solr-search-results.php" or similar. The template would just output the HTML provided by the search results.

This is similar to how other major plugins handle a default template + overrides. See [WooCommerce templating](https://docs.woocommerce.com/document/template-structure/) as an example.

PR #161, opened 2016-11-29
Add ajax to search widget

ataylorme:
Addresses #124


ataylorme:
@allan23 I see a major issue with the new Ajax functionality: only one item for each facet can be selected.

See the attached screen capture.
solr-power-facet-widget-ajax-cant-select-multiple-facets

If I search for _van_ multiple results are returned. I then facet on "Vehicle Year" but when I select a year all others disappear.

This is because there is only one year in the returned results.

I think we can solve this in two ways:
1) Return facet items not in the current search results and display them to be selected. See [this article](https://wiki.apache.org/solr/SimpleFacetParameters#Tagging_and_excluding_Filters) for more details.

2) Set a delay on fetching the updated results so the user has time to select more than one facet.

The first option is ideal and the correct solution to the problem. Please research and provide a time estimate to implement exclude filters and return all facets with the search term, even if they are in the current query.

Once we know how much work it will take we can decide if it's worth pursuing or if we can only afford the "quick fix"

includes/class-solrpower.php
```diff
@@ -195,8 +209,61 @@ function add_panel( $panels ) {
...
+			's'              => filter_input( INPUT_GET, 's', FILTER_SANITIZE_STRING ),
+			'facets'         => filter_input( INPUT_GET, 'facet', FILTER_SANITIZE_STRING, FILTER_REQUIRE_ARRAY ),
+			'posts_per_page' => get_option( 'posts_per_page' )
+		);
+
+		$query = new WP_Query( $args );
+		$query->get_posts();
+		ob_start();
+		// Check to see if an overwrite template exists in theme.
+		$theme_template = locate_template( array( 'solr-search-results.php' ), true, true );
```
ataylorme: @allan23 `includes/class-solrpower.php` has the following:
```
// If there is a template file then we use it
if ( file_exists( TEMPLATEPATH . '/s4wp_search.php' ) ) {
	// use theme file
	include_once( TEMPLATEPATH . '/s4wp_search.php' );
} else if ( file_exists( dirname( __FILE__ ) . '/template/s4wp_search.php' ) ) {
	// use plugin supplied file
	add_action( 'wp_head', array( $this, 'default_head' ) );
	include_once( dirname( __FILE__ ) . '/template/s4wp_search.php' );
} else {
	// no template files found, just continue on like normal
	// this should get to the normal WordPress search results
	return;
}
```

Is the `template/s4wp_search.php` still being used?

ataylorme: @allan23 okay, in that case let's remove it and get rid of the technical debt of carrying around legacy code.

joshkoenig:
I think there are different legit use-cases for facet UI behavior, and we should consider whether we want to settle on one, or provide some options.

For instance: it might be the case that I want to be able to select a range of model-years to browse, but it might also be the case that I want to help customers "drill down" in a product catalog. We also need to consider what should happen when there are taxonomies that allow many categories per post.

My sense is that we do want the faceting widget to be alive when AJAX is enabled. The question is what it's behavior should be out of the box, and to what extent (and how) we allow it to be customized. 

It could be kind of awesome if we could allow developers to use the standard method of altering `WP_Query()` to affect facet behavior by changing the [facet params](https://wiki.apache.org/solr/SimpleFacetParameters#Tagging_and_excluding_Filters) and have the JS simply render the results no matter what. This would give us a very practical tutorial to get people started with customizing queries.

ataylorme:
Notes from Josh's testing:
[] Pagination isn't working
[] Display excerpt in default theme

ataylorme:
Closing this out in favor of the more updated #177

Issue #164, opened 2016-12-06
Feature Request: Add a hook when indexing all posts has been completed


cjbos:
We have a few terms which we display as regular posts on our site, I've indexed them into solr as if they are posts via a cli command which is run after we re-index all the posts or load the terms. (These terms are pulled from an external source and not updated)

It would be really handle if we could have a hook of some sort once indexing all the posts has been completed so we don't have to manually run this command.

ataylorme:
@cjbos adding this hook is a good idea. We're currently working on the next release so I'm assigning this to myself to get it added in.

cjbos:
Great, thanks!

ataylorme:
@cjbos The proposed action is `solr_power_index_all_finished` and will fire when indexing all posts, either from the settings panel or wp-cli, has finished.

Note that the Solr index also updates when individual posts are modified, deleted, etc. This action only fires for the bulk index, not indexing individual posts.

cjbos:
Thanks!

ataylorme:
We still have a few other features to finish before we cut a release but once this ships I'll let you know.

PR #167, opened 2016-12-13
Add 'solr_power_index_all_finished' action

ataylorme:
Fired when indexing all posts is complete.

Addresses #164


ataylorme:
@allan23 Can you review this? Want to make sure I added the actions in the correct place.

The `solr_power_index_all_finished` action should fire after indexing all posts, either from the admn or wp-cli, is complete.

allan23:
@ataylorme this works correctly for me.

PR #169, opened 2016-12-13
Parse 's' argument when solr_integrate is true

ataylorme:
Addresses #168


ataylorme:
@allan23 please review when you can. This PR is so that the `s` argument of for `WP_Query` is used when `solr_integrate` is true.

ataylorme:
> should there be a unit test for this
@allan23 Yes, I think so. If you have time and it'll be quick go ahead and write one otherwise I can try and get to it later this week.

ataylorme:
Test added, merging this in

PR #171, opened 2016-12-14
Checks for searchable post type before indexing modified post

cjbos:


ataylorme:
@allan23 can you add this to your review list?

cjbos:
Sure, I added the strict flag to the in_array check

PR #198, opened 2017-01-26
Use the default scheme when submitting schema

danielbachhuber:


includes/class-solrpower-api.php
```diff
@@ -532,4 +515,23 @@ function dismax_query( $query, $dismax ) {
 
 		return $query;
 	}
+
+	/**
+	 * Check for the SOLR_POWER_SCHEME constant.
+	 * If it exists and is "http" or "https", use it as the default scheme value.
+	 */
+	private function get_default_scheme() {
```
ataylorme: Breaking this out into a reusable function is a good call.

includes/class-solrpower-api.php
```diff
@@ -64,7 +64,7 @@ function submit_schema() {
 		}
 
 		$path        = $this->compute_path();
-		$url         = 'https://' . getenv( 'PANTHEON_INDEX_HOST' ) . ':' . getenv( 'PANTHEON_INDEX_PORT' ) . '/' . $path;
```
ataylorme: Pantheon should always be https but that is the default value of `get_default_scheme()` so not hard coding it should be fine.

danielbachhuber: One challenge with this is that the code still requires a `$client_cert`, which isn't really applicable for `http`

Issue #191, opened 2017-01-23
Better UX for restarting index process on large site


danielbachhuber:
As a Solr Power user with 200k posts, it's painful to have the index process die unexpectedly halfway through.

It would be more enjoyable to have a better UX for restarting the index process on a large site. Couple of ideas:

* Include an `--offset` argument for the `wp solr index` command so I can restart the process after a certain number of posts.
* Create functionality where WordPress knows what posts are in the index, and thus which aren't, so it can limit the index process to only posts that aren't in the index.

danielbachhuber:
Actually, because the `paged` query var is used in `wp solr index`, we wouldn't want to support `offset` directly, because `offset` overrides `paged` (and would break the iteration).

danielbachhuber:
@ataylorme For the admin, what do you think about this?

### No index to resume

![image](https://cloud.githubusercontent.com/assets/36432/22353879/6f7b284e-e3d7-11e6-8b91-1b428758de1c.png)

### Index to resume

![image](https://cloud.githubusercontent.com/assets/36432/22353930/b7d2feb4-e3d7-11e6-965d-97aa940d8c92.png)

ataylorme:
@danielbachhuber that looks good to me. I think a total would be nice (x/x posts indexed) but not necessary.

Since there is a resume would it be tough to add pause as well?

danielbachhuber:
> Since there is a resume would it be tough to add pause as well?

It wouldn't be tough. I'm not convinced it's necessary, though.

PR #201, opened 2017-01-26
Make Solr indexing resumable by default for CLI command

danielbachhuber:
With this change, `wp solr index` keeps track of which page it's gotten
to in the index. If the index is interrupted, it will resume where it
was interrupted, unless the `--force` flag is used.

See #191

### DIFF
--- a/includes/class-solrpower-cli.php
+++ b/includes/class-solrpower-cli.php
@@ -1,5 +1,7 @@
 <?php
 
+use WP_CLI\Utils;
+
 /**
  * Perform a variety of actions against your Solr instance.
  */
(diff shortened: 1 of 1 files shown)


includes/class-solrpower-cli.php
```diff
@@ -79,20 +84,39 @@ public function delete( $args, $assoc_args ) {
...
+			'orderby'        => 'ID',
+			'post_type'      => get_post_types( array( 'exclude_from_search' => false ) ),
 		);
 		// Check if specified post_type is valid.
 		if ( isset( $assoc_args[ 'post_type' ] ) && (false === post_type_exists( $assoc_args[ 'post_type' ] )) ) {
 			WP_CLI::error( '"' . $assoc_args[ 'post_type' ] . '" is an invalid post type.' );
 		}
 		$query_args		 = array_merge( $defaults, $assoc_args );
+
+		$log_key         = 'solr_index_paged_' . md5( serialize( $query_args ) );
```
danielbachhuber: I discovered this logic isn't quite right, because `$query_args` contains `--force`, which means a resume invocation without `--force` won't produce the same log key.

danielbachhuber:
Closing in favor of #204

PR #204, opened 2017-02-15
Refactor batch indexing to be resumeable

### LINKED ISSUE #191: Better UX for restarting index process on large site

danielbachhuber:
* [x] Introduce new `SolrPower_Batch_Index` class for handling batch indexing.
* [x] Capture and resume index based off `paged`
* [x] Refactor `wp solr index` to use batch index class and provide more verbosity.
* [x] Refactor admin UX to use batch index class.
* [x] Remove old `wp solr index` command.
* [x] Remove old code associated with admin indexing UX.
* [ ] Add tests around the batch index process.
* [x] Fire the `solr_power_index_all_finished` action at the end of the indexing process.

Fixes #191


danielbachhuber:
> However, the admin UI hung with 81 posts left.

Ah, probably because I still need to:

image

ataylorme:
Screenshots

screen shot 2017-02-23 at 11 03 11 am
screen shot 2017-02-23 at 11 10 20 am

ataylorme:
Hah! I totally missed that. Once that's done I'll test again and approve the PR.

includes/class-solrpower-batch-index.php
```diff
@@ -0,0 +1,265 @@
...
+		return false;
+	}
+
+	/**
+	 * Increment value for 'paged', and update its value in the cache
+	 */
+	public function increment_page() {
+		$this->query_args['paged'] = $this->query_args['paged'] + 1;
+		delete_option( $this->batch_cache_key );
+		add_option( $this->batch_cache_key, $this->query_args['paged'], null, 'off' );
```
danielbachhuber: > > Are you using `delete_option()` and `add_option()` rather than `update_option()` to disable autoloading in cases where the option doesn't exist and needs to be added?

Yes

> If so the `$autoload` parameter should be `false` or `'no'`, not `'off'`

D'oh! Fixed bd036ac

includes/class-solrpower-batch-index.php
```diff
@@ -0,0 +1,265 @@
...
+		return $this->failed_posts;
+	}
+
+	/**
+	 * Whether or not there are remaining posts to index.
+	 *
+	 * @return bool
+	 */
+	public function have_posts() {
+		if ( ! empty( $this->post_ids ) ) {
```
danielbachhuber: Sure, 911064a

includes/class-solrpower-batch-index.php
```diff
@@ -0,0 +1,265 @@
...
+		// Out of posts, so remove our offset.
+		delete_option( $this->batch_cache_key );
+		return false;
+	}
+
+	/**
+	 * Increment value for 'paged', and update its value in the cache
+	 */
+	public function increment_page() {
+		$this->query_args['paged'] = $this->query_args['paged'] + 1;
```
danielbachhuber: Sounds good, 60252a2

danielbachhuber:
> Hah! I totally missed that. Once that's done I'll test again and approve the PR.

I've updated the UI for indexing, and added a completion message:

indexing-ux

I still want to add tests for the batch index class, but can do so in a follow-up PR #217

This PR is good to 🚢

ataylorme:
@danielbachhuber this is looking good to me. @joshkoenig any objection to merging?

Issue #203, opened 2017-02-03
Solr exception when trying to index CPT with custom meta 'year'


danielbachhuber:
```
Feb 03, 2017 3:57:52 PM org.apache.solr.common.SolrException log
SEVERE: org.apache.solr.common.SolrException: ERROR: [doc=5339] multiple values encountered for non multiValued field year_i: [2017, 1998]
```

danielbachhuber:
Solution https://github.com/danielbachhuber/solr-demo-redux/commit/45e04faa8bad870507dad65adb36f633b9e6fab5

danielbachhuber:
This was also ultimately necessary https://github.com/danielbachhuber/solr-demo-redux/commit/79489c375dcdad398a803a8efbc8d34a8f002e56

PR #209, opened 2017-02-17
Apply a filter to the generated Solr document

danielbachhuber:
This filter permits third-party code to massage the automatically
generated document as needed.

See #203


danielbachhuber:
Actually, going to land this as-is to use on #204

Issue #215, opened 2017-02-23
WP admin search defaults to solr


aayler:
We recently upgraded to WP core 4.7.2 (from 4.6.1) and additionally upgraded solr-power to 1.0 from 0.6.

Following this, we're seeing that, without setting the "solr-integrate" parameter to true or using the "solr_allow_admin" filter hook, searches in our wp-admin (of custom post types, for instance) are using Sol, and returning thousands of results rather than the expect few ... any ideas on why this is happening and how to disable this?

And when we revert back to solr-power 0.6, the search returns to normal.

Thanks!

ataylorme:
> searches in our wp-admin (of custom post types, for instance) are using Sol, and returning thousands of results rather than the expect few

@aayler So the front-end of the site is returning results as expected and the issue is only in the admin?

Just want to make sure I try reproduce in the appropriate place.

![screen shot 2017-02-23 at 4 11 40 pm](https://cloud.githubusercontent.com/assets/2133004/23281430/e54fa42a-f9e2-11e6-8c62-42d46ba5711d.png)

aayler:
hi @ataylorme, i appreciate the prompt response!

unfortunately, the site search is off as well, in terms of the weights (boost) we've placed on search results. in other words, in the frontend, the results are correct but not in the desired display order.

and yes, i am referring to the search in the admin highlighted by the red arrow in your image.

(and we've reindexed the site, which is giving - as expected slightly different, likely new content, results, but the same behavior)

thanks again!

ataylorme:
> unfortunately, the site search is off as well, in terms of the weights (boost) we've placed on search results. in other words, in the frontend, the results are correct but not in the desired display order.

Can you open a separate issue for that? Including details on what type of content you are indexing, which boosts are being used, etc. will  be useful.

> and yes, i am referring to the search in the admin highlighted by the red arrow in your image.
Thanks for confirming. I'll try to reproduce.

Currently I would advise you to stick on `0.6` in production as we diagnose these issues. If you have another environment, like dev, you can work with me on these issues in that would be ideal.

aayler:
@ataylorme : thanks, will do re: site search issues.

we haven't pushed solr-power 1.0 to *live* yet, and in fact, we are working on a multidev testing env on pantheon ... what would be the best way to get in touch with you re: the site URL and credentials for you to review?

thanks again,

ataylorme:
> we haven't pushed solr-power 1.0 to live yet

[...]

PR #216, opened 2017-02-23
Disallow Solr in WordPress admin unless coming from admin ajax

ataylorme:
Addresses #215

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -81,7 +81,7 @@ class SolrPower_WP_Query {
 		 *
 		 * @param bool $solr_allow_admin True to query in WordPress Dashboard or false [default false].
 		 */
-		if ( is_admin() && false === apply_filters( 'solr_allow_admin', false ) && (!defined( 'DOING_AJAX' ) || !DOING_AJAX ) ) {
+		if ( is_admin() && false === apply_filters( 'solr_allow_admin', false ) && ( ! ( defined( 'DOING_AJAX' ) && DOING_AJAX ) ) ) {
 			return;
 		}
 
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -81,8 +81,8 @@ class SolrPower_WP_Query {
 		 *
 		 * @param bool $solr_allow_admin True to query in WordPress Dashboard or false [default false].
 		 */
-		if ( is_admin() && false === apply_filters( 'solr_allow_admin', false ) ) {
-			// return;
+		if ( is_admin() && false === apply_filters( 'solr_allow_admin', false ) && (!defined( 'DOING_AJAX' ) || !DOING_AJAX ) ) {
+			return;
 		}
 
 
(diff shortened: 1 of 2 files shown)


ataylorme:
@joshkoenig or @danielbachhuber mind giving this a review? See #215 for background

includes/class-solrpower.php
```diff
@@ -251,8 +255,8 @@ function ajax_search() {
 		} );
 
 		// Allow an AJAX search.
-		add_filter( 'solr_allow_ajax', '__return_true' );
-		add_filter( 'solr_allow_admin', '__return_true' );
+		add_filter( 'solr_allow_ajax', array( $this, 'return_true' ) );
+		add_filter( 'solr_allow_admin', array( $this, 'return_true' ) );
```
danielbachhuber: Why create a dedicated method for this?

includes/class-solrpower-wp-query.php
```diff
@@ -81,8 +81,8 @@ function setup() {
 		 *
 		 * @param bool $solr_allow_admin True to query in WordPress Dashboard or false [default false].
 		 */
-		if ( is_admin() && false === apply_filters( 'solr_allow_admin', false ) ) {
-			// return;
+		if ( is_admin() && false === apply_filters( 'solr_allow_admin', false ) && (!defined( 'DOING_AJAX' ) || !DOING_AJAX ) ) {
```
danielbachhuber: This:

```
(!defined( 'DOING_AJAX' ) || !DOING_AJAX )
```

Would be better expressed as:

```
! ( defined( 'DOING_AJAX' ) && DOING_AJAX )
```

ataylorme:
> My only question is whether we have other use-cases where we want to allow solr power to take over on the backend. If not, wouldn't it be simpler just to disengage under any circumstance where is_admin() is true?

@joshkoenig there's an existing filter for `solr_allow_admin` which defaults to `false`. I see no harm in leaving it for backwards compatibility. It also gives us the option to do Solr in the admin correctly in the future.

includes/class-solrpower.php
ataylorme: @danielbachhuber it was the end of the day and I brain farted thinking about anonymous functions and how they are a pain to remove once added. `__return_true` is a standard PHP function so you're right we should be using it.

includes/class-solrpower-wp-query.php
ataylorme: Fixed in the latest commit

ataylorme:
@joshkoenig since Daniel is out on vacation can you approve the PR changes? I addressed the items in his code review and would like to merge.

includes/class-solrpower.php
```diff
@@ -250,9 +250,9 @@ function ajax_search() {
 			return $url;
 		} );
 
-		// Allow an AJAX search.
-		add_filter( 'solr_allow_ajax', '__return_true' );
-		add_filter( 'solr_allow_admin', '__return_true' );
+		// Add __return_true to the admin/ajax filters so the widget can access admin ajax.
+		add_filter( 'solr_allow_ajax', array( $this, '__return_true' ) );
+		add_filter( 'solr_allow_admin', array( $this, '__return_true' ) );
```
danielbachhuber: Not quite correct either. Should be:

```
add_filter( 'solr_allow_ajax', '__return_true' );
```

This should've failed the tests, so it'd be good to add a test that fails with this invalid callback.

ataylorme: d'oh - forgot to remove when I removed the class method.

ataylorme: @danielbachhuber can you review one more time?

Issue #218, opened 2017-02-24
Boosting hard-coded


ataylorme:
> hi @ataylorme : regarding the front end issue, we located the issue (for us) = https://github.com/pantheon-systems/solr-power/blob/1.0.0/includes/class-solrpower-api.php#L359 this call dramatically changes the order of our search results ...

> is there a way to override or reset/unset those "hard-coded" values, ie, 'post_title^25 post_content^50'? (shouldn't boosting be a functionality open to use case?)

> and a quick question re: that line -- on = http://solarium.readthedocs.io/en/stable/queries/select-query/building-a-select-query/components/dismax-component/, the description of the boostFunctions method describes the string composed in the format: "funcA(arg1,arg2)^1.2 funcB(arg3,arg4)^2.2". are post_title and post_content references to functions?

`post_title` and `post_content` boosts are hard-coded. Boosts should, at a minimum, be filterable.

See [class-solrpower-api.php#L359](https://github.com/pantheon-systems/solr-power/blob/1.0.0/includes/class-solrpower-api.php#L359)

@aayler can you provide more details on how you are boosting?

aayler:
Hi @ataylorme : we're using the setQueryFields method in the solr_query filter handler.

so something like = `$dismax->setQueryFields('post_title^2.5 tags^1.5');`

ataylorme:
Just FYI the reason we set a boost to begin with is that when facets are used search results weren't really relevant. Adding a boost to title and content helped keep them relevant but _what_ is being boosted should be filterable.

ataylorme:
@aayler can you test the changes in [946e0b12096a10d22bf3d56fd69cf24896895faa](https://github.com/pantheon-systems/solr-power/pull/219/commits/946e0b12096a10d22bf3d56fd69cf24896895faa) on your multidev?

You should be able to use the `solr_boost_query` filter to override the boost value of `post_title^25 post_content^50`.

ataylorme:
Make that the [latest changes in the PR](https://github.com/pantheon-systems/solr-power/pull/219/files)

aayler:
@ataylorme 👍  on our end re: filtering the `solr_boost_query`

thanks again!

ataylorme:
@aayler great! I'll ask for a code review and merge this in for `1.1`. Once we tag `1.1` I'll ping you for testing if that's okay.

aayler:
@ataylorme : absolutely, please do; would be glad to help test.

[...]

PR #219, opened 2017-03-01
Make boost query filterable

ataylorme:
Addresses #218


ataylorme:
Travis logs are failing to load making this hard to determine what's failing.

Issue #221, opened 2017-03-02
Admin Options Not Saving


ataylorme:
Error message `Action failed. Please try again.solr_update` being received from the action tab.

PR #222, opened 2017-03-02
Remove unused update case

ataylorme:
Addresses #221 

See [class-solrpower-options.php#L278](https://github.com/pantheon-systems/solr-power/blob/4ec4b5e5a69e93a4fb1e011201e7a84904f7332a/includes/class-solrpower-options.php#L278)

I don't see the `update` case being used, going to remove it.


ataylorme:
Closes #221

Issue #207, opened 2017-02-17
SolrPower_Api::get_instance()->last_error->getMessage() can include poor output


danielbachhuber:
It would be nice if we pre-processed this HTML in some way:

```
Failed to index '1 Contra Todos' (37): Solr HTTP error: OK (400)
&lt;html&gt;
&lt;head&gt;
&lt;meta http-equiv=&quot;Content-Type&quot; content=&quot;text/html; charset=ISO-8859-1&quot;/&gt;
&lt;title&gt;Error 400 ERROR: [doc=37] multiple values encountered for non multiValued field year_i: [2017, 2016]&lt;/title&gt;
&lt;/head&gt;
&lt;body&gt;&lt;h2&gt;HTTP ERROR 400&lt;/h2&gt;
&lt;p&gt;Problem accessing /solr/update. Reason:
&lt;pre&gt;    ERROR: [doc=37] multiple values encountered for non multiValued field year_i: [2017, 2016]&lt;/pre&gt;&lt;/p&gt;&lt;hr /&gt;&lt;i&gt;&lt;small&gt;Powered by Jetty://&lt;/small&gt;&lt;/i&gt;&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;
&lt;br/&gt;

&lt;/body&gt;
&lt;/html&gt;
```

From https://github.com/pantheon-systems/solr-power/issues/202#issuecomment-280735004

PR #227, opened 2017-03-20
Pull out error message document title in `SolrPower_Batch_Index`

danielbachhuber:
Solr can return full HTML in its exception, which is then escaped by
`SolrPower_Sync`. If the batch indexer detects a full HTML error, the
`<title>` is a good summary to display instead.

See #207

### DIFF
--- a/includes/class-solrpower-batch-index.php
+++ b/includes/class-solrpower-batch-index.php
@@ -228,7 +228,12 @@ class SolrPower_Batch_Index {
 		} else {
 			$this->failed_posts++;
 			$result['status'] = 'failed';
-			$result['message'] = SolrPower_Sync::get_instance()->error_msg;
+			$error_msg = SolrPower_Sync::get_instance()->error_msg;
+			// Pull out the document <title> when this looks like escaped HTML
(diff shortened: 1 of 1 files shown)


includes/class-solrpower-batch-index.php
```diff
@@ -228,7 +228,12 @@ public function index_post() {
 		} else {
 			$this->failed_posts++;
 			$result['status'] = 'failed';
-			$result['message'] = SolrPower_Sync::get_instance()->error_msg;
+			$error_msg = SolrPower_Sync::get_instance()->error_msg;
+			// Pull out the document <title> when this looks like escaped HTML
+			if ( preg_match( '#' . preg_quote( '&lt;title&gt;' ) . '(.+)' . preg_quote( '&lt;/title&gt;' ) . '#', $error_msg, $matches ) ) {
```
ataylorme: @danielbachhuber could `'(.+)'` be `'([^&]+)'` instead? I might be overly cautious but am weary of matching `.` with regex.

danielbachhuber: > could `'(.+)'` be `([^&]+)'` instead?

Probably not. If `<title>` contained `&`, then we'd miss part of the error message.

> I might be overly cautious but am weary of matching `.` with regex.

It should be fine. As long as the end sequence exists, `.` will only match to the end sequence.

Issue #41, opened 2015-12-29
Multisite Indexing


allan23:
How should indexing be handled in a multisite environment?
Should each site have their own admin page or just one main page for the network admin?

CC: @joshkoenig

joshkoenig:
This is a good question.

I do not believe that having different configuration or operations (e.g. what post types to index, index reset, etc) for individual sites in a network is a good idea. Solr doesn't have this notion built into itself, and so we shouldn't present a false set of options. There's still only once `schema.xml` for all the content, no good way to get stats for a particular site. 

We should set the assumption explicitly that this will not work out well if you're abusing a multisite setup to run a bunch of divergent sites off one install. 

So, it's one admin interface for the network admin.

I do think we will need to have some configuration options, ideally with the ability to hard-code with a `CONSTANT`(ala the other core site network config settings).

I can see the following scenarios:

1. Searches are not at all segmented meaning search is "network-wide"
2. All searches are automatically scoped to the site from which they launched via a facet. No way to search "network-wide".
3. Sites in-network are available as an optional facet (possibly a good one to consider for #37), meaning a user can limit their search to a set of sites in the network if they choose.

joshkoenig:
Let's consider this for the next release.

ataylorme:
I think we should consider resolving the existing issues and feature requests first and tackle multisite compatibility after a version 1.0 release.

taija:
We are moving several large wordpress multisite installations to Pantheon soon, and would love to be able to use Solr. Fingers crossed for multisite support soon!

ataylorme:
@danielbachhuber this would be a good candidate for work either before or after debug bar support.

@taija can you expand on how you plan on using Solr Power in those multisites?
* Will you be using search (native of the widget), `WP_Query` support or both? Which features of these items are important?
* What type of content will you be indexing?
    - post types
    - meta data
    - taxonomy data

taija:
@ataylorme - we would be using it as a drop-in replacement for WP search (native). 

We would like to index post types and taxonomy data, in that priority order.

ataylorme:
@taija regarding:

> 1. Searches are not at all segmented meaning search is "network-wide"
> 2. All searches are automatically scoped to the site from which they launched via a facet. No way to search "network-wide".

As a user which would you expect to be the default?

taija:
@ataylorme - for us, we have built the expectation that search happens network-wide, as this works well for the type of content we have.

[...]

PR #228, opened 2017-03-20
Permit Solr Power to be used in WordPress multisite

danielbachhuber:
* [x] When active on multisite, always use network admin settings.
* [x] Always include `blogid` in the post document sent to Solr.
* [x] Use `blogid` as a facet when overloading `WP_Query`.
* [ ] Indexing UI TK
* [ ] Include tests covering multisite integration

See #41


danielbachhuber:
Interestingly, the test failure seems to happen to the last run test:

image

ataylorme:
@danielbachhuber have you merged the latest changes from `master`?

danielbachhuber:
> have you merged the latest changes from master

Yes, just did. No impact. I'll debug some more when I have a moment.

includes/class-solrpower-wp-query.php
```diff
@@ -418,7 +418,11 @@ private function build_query( $query ) {
 			'name'    => 'post_name'
 		);
 		if ( ! $query->get( 'solr_integrate' ) ) {
-			return $query->get( 's' );
+			$s = $query->get( 's' );
+			if ( is_multisite() ) {
+				$s .= ' AND(blogid:' . get_current_blog_id() . ')';
+			}
```
danielbachhuber: @ataylorme Any ideas why this would cause the test to fail? It's pretty much destroying me.

Without this conditional, the test passes but my test site displays two posts:

image

With the conditional, the test fails, but my site correctly displays one post.

The query for the failing Solr test is:

```
[10-Apr-2017 23:05:00 UTC] array (
  'handler' => 'select',
  'resultclass' => 'Solarium\\QueryType\\Select\\Result\\Result',
  'documentclass' => 'Solarium\\QueryType\\Select\\Result\\Document',
  'query' => 'Movie Reviews AND(blogid:1)',
  'start' => 0,
  'rows' => 10,
  'fields' => '*,score',
  'omitheader' => false,
  'sort' => 
  array (
    'score' => 'desc',
  ),
  'querydefaultoperator' => 'OR',
)
```

The query for the correct search results is:

```
[10-Apr-2017 23:05:33 UTC] array (
  'handler' => 'select',
  'resultclass' => 'Solarium\\QueryType\\Select\\Result\\Result',
  'documentclass' => 'Solarium\\QueryType\\Select\\Result\\Document',
  'query' => 'foo AND(blogid:2)',
  'start' => 0,
  'rows' => 10,
  'fields' => '*,score',
  'omitheader' => false,
  'sort' => 
  array (
    'score' => 'desc',
  ),
  'querydefaultoperator' => 'OR',
)
```

includes/class-solrpower-sync.php
```diff
@@ -82,12 +82,7 @@ function handle_delete( $post_id ) {
 		$delete_page          = $plugin_s4wp_settings['s4wp_delete_page'];
 		$delete_post          = $plugin_s4wp_settings['s4wp_delete_post'];
 
-
-		if ( is_multisite() ) {
-			$this->delete( get_current_blog_id() . '_' . $post_info->ID );
-		} else {
-			$this->delete( $post_info->ID );
-		}
+		$this->delete( get_current_blog_id() . '_' . $post_info->ID );
```
danielbachhuber: @ataylorme This is potentially problematic, actually.

If we leave this change in, this means existing documents in the index will be orphaned. They'll be orphaned because WordPress thinks their `solr_id` is `{$blog_id}_{$post_id}`, but Solr knows the documents as `{$post_id}`

I think we need to walk back this change change to the document `solr_id` structure:

image

Instead, we'll need some note that the user needs to re-index if they've switched a WordPress install from single site to multisite.

danielbachhuber:
Closing in favor of #243

Issue #229, opened 2017-03-20
Consolidate calls to get_post_types()


danielbachhuber:
The following is called five times in the codebase:

```
apply_filters( 'solr_post_types', get_post_types( array( 'exclude_from_search' => false ) ) )
```

Instead, it would be better to create `SolrPower::get_post_types()`, and have each instance call this method directly.

ataylorme:
Closed via #234

PR #234, opened 2017-04-07
Post status filter

ataylorme:
* Add get_post_types and get_post_statuses helper methods; Addresses #229
* Remove unnecessary get_env call
* Add solr_post_status filter
* Add missing solr_post_types filter to get_post_types call
* Use PANTHEON_INDEX_PORT instead of hard-coded port in curl

### DIFF
--- a/includes/class-solrpower-api.php
+++ b/includes/class-solrpower-api.php
@@ -432,7 +432,7 @@ class SolrPower_Api {
 		$stats     = wp_cache_get( $cache_key, 'solr' );
 		if ( false === $stats ) {
 
-			$post_types = get_post_types( array( 'exclude_from_search' => false ) );
+			$post_types = apply_filters('solr_post_types', get_post_types( array( 'exclude_from_search' => false ) ) );
 
 			$stats = array();
 			foreach ( $post_types as $type ) {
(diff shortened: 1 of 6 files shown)


includes/class-solrpower-api.php
```diff
@@ -432,7 +432,7 @@ function index_stats() {
 		$stats     = wp_cache_get( $cache_key, 'solr' );
 		if ( false === $stats ) {
 
-			$post_types = get_post_types( array( 'exclude_from_search' => false ) );
+			$post_types = apply_filters('solr_post_types', get_post_types( array( 'exclude_from_search' => false ) ) );
```
danielbachhuber: Should this use the filter too?

includes/class-solrpower.php
```diff
@@ -313,4 +313,22 @@ function ajax_search() {
...
+	}
+
+	/**
+	 * Returns the post statuses used for Solr after applying the solr_post_status filter
+	 *
+	 * @return array
+	 */
+	public function get_post_statuses() {
+		return apply_filters( 'solr_post_status', array( 'publish' ) );
+	}
```
danielbachhuber: Can we make these static methods, so they don't need to be called with the instance?

ataylorme: Good call, updated.

includes/class-solrpower-api.php
ataylorme: 👍

ataylorme:
@danielbachhuber can you review again?

gtenaschuk:
Great! Going to update my repo.
I was not available today to make the changes, thanks guys!

ataylorme:
Thank you for submitting the PR to begin with!

PR #246, opened 2017-04-17
Respect all query vars when searching with WP_Query

### LINKED ISSUE #244: WP Query ignores post_type query_var
When using the plugin with the default WordPress query it ignores restricting results by post_type. I've tested this on the WordPress search page using `?s=my-term&post_type=my-post-type`. Regardless of the post type set the query returns all post types.

danielbachhuber:
The prior behavior would bail early when `s` was set but
`solr_integrate` wasn't provided, which meant query vars would be
discarded when searching

Fixes #244

### DIFF
--- a/tests/phpunit/wp_query/test-wp-query.php
+++ b/tests/phpunit/wp_query/test-wp-query.php
@@ -160,4 +160,16 @@ class SolrWPQueryTest extends SolrTestBase {
 		wp_reset_postdata();
 		SolrPower_Api::get_instance()->ping=true;
 	}
+
+	public function test_wp_query_search_filter_post_type() {
+		$post_id = $this->__create_test_post();
+		$page_id = $this->__create_test_post( 'page' );
+		$args = array(
+			's'         => 'Test',
+			'post_type' => 'page',
+		);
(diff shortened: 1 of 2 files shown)


tests/phpunit/wp_query/test-wp-query.php
```diff
@@ -160,4 +160,16 @@ function test_wp_query_failed_ping() {
...
+	public function test_wp_query_search_filter_post_type() {
+		$post_id = $this->__create_test_post();
+		$page_id = $this->__create_test_post( 'page' );
+		$args = array(
+			's'         => 'Test',
+			'post_type' => 'page',
+		);
+		$query = new WP_Query( $args );
+		$this->assertEquals( 1, $query->post_count );
+		$this->assertEquals( 1, $query->found_posts );
```
ataylorme: @danielbachhuber let's also assert that the post ID matches `$page_id` and that the post type of the resulting query is, indeed, `page`.

danielbachhuber: Updated in 319ec15

PR #247, opened 2017-04-19
Remove duplicate options when initializing them

gtenaschuk:
- Remove options initialized twice


danielbachhuber:
Thanks @gtenaschuk !

PR #249, opened 2017-04-19
Add missing filter for custom fields

gtenaschuk:
Custom fields are not being included in search query, also part of the code was replaced in solrpower sync file since the `s4wp_index_custom_fields` options is parsed as an array when options are being loaded and it will just validate that the value is an array before using it.

Fixes and improvements
- Add missing filter for custom fields when searching
- Add custom fields validation when parsing meta query


gtenaschuk:
Doing it now @danielbachhuber

danielbachhuber:
> I added the test for my changes

Thanks!

> but an existing one `test_wp_query_failed_ping` failed. It is weird since it pass in my local environment.

Known issue that we've been ignoring. I've filed an issue to actually look into it: #274

Issue #248, opened 2017-04-19
Searching by author name?


mindofjonas:
How would I implement a search for posts by author name using Solr Power? When using a default WP Query like`s=author-name` it will return both posts by that post_author and posts where the author's name appears in the post_title/post_content. However Solr Power returns only the later.

ataylorme:
> However Solr Power returns only the later.

@mindofjonas are you not seeing the results at all or are they lower in the search results?

Solr Power boosts `post_title` and `post_content`. See [`includes/class-solrpower-api.php`](https://github.com/pantheon-systems/solr-power/blob/master/includes/class-solrpower-api.php#L355:L365). 

The boosted items are filterable via [`solr_boost_query`](https://github.com/pantheon-systems/solr-power/wiki/Filters#solr_boost_query).

`WP_Query` with Solr also has support for `post_author` as a parameter, rather than including it in the search. 

Author is also a facet available to the facet search widget.

mindofjonas:
@ataylorme it returns only posts where the author's name is explicitly included in the title or content.

I've tried using the boost filter as below but that doesn't return results by an author. 
```
function wlw_solr_query( $dismax, $query ){
    $dismax->setQueryFields('post_author^2.5 post_title^2 post_content^1.5');
    return $dismax;
}
add_action( 'solr_dismax_query', 'wlw_solr_query', 10, 2 );
```
We were originally using a [filter](https://danielbachhuber.com/2012/02/07/include-posts-by-matching-authors-in-your-search-results/) on `posts_search` (credit goes to @danielbachhuber) that modifies the SQL run during a search query so that a search for 'author name' would return posts by that author.

I'd like to use something similar to the above WP Query behavior instead of relying on using `post_author=author-name` or a facet on the frontend as that would add complexity to our search page.

My initial thought is to use an instance of the`SolrPower_Api` class with a modified`$qry` that also searches the `post_author` field. However I don't have experience with Solarium

danielbachhuber:
> it returns only posts where the author's name is explicitly included in the title or content.
>
> I've tried using the boost filter as below but that doesn't return results by an author.

I was able to get this to work, without the boost filter, by modifying `schema.xml` to include:

```
<copyField source="post_author" dest="text"/>
```

See #250

> We were originally using a filter on posts_search (credit goes to @danielbachhuber) that modifies the SQL run during a search query so that a search for 'author name' would return posts by that author.

[...]

PR #250, opened 2017-04-19
Match author name in search results

danielbachhuber:
See #248

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,16 @@ A meta_query can use the following compare operators:
 
 (```'REGEXP'```, ```'NOT REGEXP'```, and ```'RLIKE'``` are not supported.)
 
+## Configuration Tips ##
+
+### Searching by author name ###
+
+To support searching by author name (e.g. where "Pantheon" would return posts authored by the "Pantheon" user), add the following to your custom `schema.xml`:
+
+```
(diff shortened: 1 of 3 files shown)


danielbachhuber:
@ataylorme Do you think it makes sense to include as default behavior, or is it something we should just have documented?

ataylorme:
@danielbachhuber I think it makes sense to match WordPress default behaviour. I'm fine with adding this since `post_title` and `post_content` are still boosted. For example, if I had a post titles `Andrew Jackson` it would return in the results above an author match for `Andrew`

danielbachhuber:
> I think it makes sense to match WordPress default behavior. I'm fine with adding this since `post_title` and `post_content` are still boosted.

WordPress' default behavior is that author name isn't used to produce search results. These two sentences are contradicting statements...

ataylorme:
Sorry, I was reading the issue description in #248.

Documentation is fine. Something like "To have Solr Power match the search term to additional fields besides `post_title` and `post_content` add them to the schema. Example:"

danielbachhuber:
> Documentation is fine. Something like "To have Solr Power match the search term to additional fields besides post_title and post_content add them to the schema.

Sounds good, updated.

PR #262, opened 2017-04-27
Enable WPCS on templates/*

danielbachhuber:
See #85


danielbachhuber:
Circle failure is unrelated — `npm install` is now failing.

Issue #270, opened 2017-04-27
Replicate WordPress' default search ordering behavior


danielbachhuber:
In a stock WordPress install, the default behavior is `orderby=relevance`. This is then interpreted to:

```
if ( $q['search_terms_count'] > 1 ) {
	$num_terms = count( $q['search_orderby_title'] );

	// If the search terms contain negative queries, don't bother ordering by sentence matches.
	$like = '';
	if ( ! preg_match( '/(?:\s|^)\-/', $q['s'] ) ) {
		$like = '%' . $wpdb->esc_like( $q['s'] ) . '%';
	}
	$search_orderby = '';
	// sentence match in 'post_title'
	if ( $like ) {
		$search_orderby .= $wpdb->prepare( "WHEN {$wpdb->posts}.post_title LIKE %s THEN 1 ", $like );
	}
	// sanity limit, sort as sentence when more than 6 terms
	// (few searches are longer than 6 terms and most titles are not)
	if ( $num_terms < 7 ) {
		// all words in title
		$search_orderby .= 'WHEN ' . implode( ' AND ', $q['search_orderby_title'] ) . ' THEN 2 ';
		// any word in title, not needed when $num_terms == 1
		if ( $num_terms > 1 )
			$search_orderby .= 'WHEN ' . implode( ' OR ', $q['search_orderby_title'] ) . ' THEN 3 ';
	}
	// Sentence match in 'post_content' and 'post_excerpt'.
	if ( $like ) {
		$search_orderby .= $wpdb->prepare( "WHEN {$wpdb->posts}.post_excerpt LIKE %s THEN 4 ", $like );
		$search_orderby .= $wpdb->prepare( "WHEN {$wpdb->posts}.post_content LIKE %s THEN 5 ", $like );
	}
	if ( $search_orderby ) {
		$search_orderby = '(CASE ' . $search_orderby . 'ELSE 6 END)';
	}
} else {
	// single word or sentence search
	$search_orderby = reset( $q['search_orderby_title'] ) . ' DESC';
}
```

We should see if we can replicate this behavior in Solr, to make relevance ordering more consistent between Solr and native MySQL.

danielbachhuber:
https://core.trac.wordpress.org/ticket/7394 is the original ticket.

danielbachhuber:
@joshkoenig @ataylorme In #271, I have a solution that I don't love. I'm using the boost query feature of dismax to boost documents where the `post_title` has an exact string match.

[...]

PR #271, opened 2017-04-28
Boost exact match on post_title to simulate WordPress

danielbachhuber:
See #270


danielbachhuber:
Closing in favor of #272

PR #278, opened 2017-05-17
Switch to Terminus 1.x in Circle tests

### LINKED ISSUE #192: Switch to Terminus 1.0 in CI scripts
Follow the pattern in https://github.com/lcache/wp-lcache/pull/110

@tessak22 and I are going to work on this at Agency Office Hours on Wednesday: https://pantheon.io/agencies/office-hours

danielbachhuber:
Fixes #192

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,6 +5,7 @@ machine:
   environment:
     TERMINUS_ENV: ci-$CIRCLE_BUILD_NUM
     TERMINUS_SITE: solr-power
+    SITE_ENV: $TERMINUS_SITE.$TERMINUS_ENV
     WORDPRESS_ADMIN_USERNAME: pantheon
     # A random password is set in the dependencies:pre stage as a text file.
     # This line reads the same file repeatedly. If the openssl call were used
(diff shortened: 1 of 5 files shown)


circle.yml
```diff
@@ -33,9 +36,15 @@ dependencies:
...
   override:
-    - composer global require pantheon-systems/terminus ">=0.13.3 <0.99"
-    - PATH=~/.composer/vendor/bin:$PATH
-    - terminus auth login --machine-token=$TERMINUS_TOKEN
+    - |
+      if [ -z "$TERMINUS_TOKEN" ]; then
+        echo "TERMINUS_TOKEN environment variables missing; assuming unauthenticated build"
+        exit 0
+      fi
+      composer install
```
ataylorme: @danielbachhuber let's move this to a script - something like [install_terminus.sh](https://github.com/ataylorme/Advanced-WordPress-on-Pantheon/blob/master/bin/install_terminus.sh) and cache the Terminus directory so it doesn't need to be downloaded on each run.

danielbachhuber: > cache the Terminus directory so it doesn't need to be downloaded on each run.

Does it make sense to cache the Terminus directory if we're installing `master`?

I think it only makes sense to cache the Terminus directory if we're installing a specific version. Otherwise, the tests will be running against a random, unknown Terminus version.

ataylorme: I suppose not. Either way I think we can keep the `circle.yml` file a little trimmer by moving this Terminus installation into a script. If it's done in `dependencies: pre:` then we can authenticate there as well.

danielbachhuber: Sounds good. Updated in 900a2d32fd04d95b863320cc06021284d084299b

PR #281, opened 2017-05-18
Clarify error message when environment variables aren't set

### LINKED ISSUE #240: Clarify "Please enable Solr" error message on Pantheon
image

When we're in the Pantheon environment, it would be more helpful to say "Please enable the Solr add-on".

danielbachhuber:
Fixes #240


danielbachhuber:
When `PANTHEON_ENVIRONMENT` is set, the error message looks like this:

image

PR #282, opened 2017-05-18
Error early when `PANTHEON_ENVIRONMENT` isn't set

### LINKED ISSUE #230: Inconsistent user experience for submitting a schema
When I repost schema.xml through the WordPress admin, I see:

image

When I repost schema.xml through WP-CLI, I see:

```
$ wp solr repost-schema
Error: Schema repositing only works in a Pantheon environment.
```

In both cases, I'd expect to be able to repost my schema.xml. If it's not technically feasible, I'd expect to see some better error message communicating what I should do instead.

danielbachhuber:
Schema posting only works on Pantheon

Fixes #230

### DIFF
--- a/includes/class-solrpower-options.php
+++ b/includes/class-solrpower-options.php
@@ -355,7 +355,7 @@ class SolrPower_Options {
 					return;
 				}
 				if ( ! getenv( 'PANTHEON_ENVIRONMENT' ) || ! getenv( 'FILEMOUNT' ) ) {
-					$this->msg = esc_html__( 'Schema repositing only works in a Pantheon environment.', 'solr-for-wordpress-on-pantheon' );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Looks like this: 
image

includes/class-solrpower-options.php
```diff
@@ -354,6 +354,10 @@ function check_for_actions() {
 				if ( ! $this->check_nonce( 'solr_repost_schema' ) ) {
 					return;
 				}
+				if ( ! getenv( 'PANTHEON_ENVIRONMENT' ) || ! getenv( 'FILEMOUNT' ) ) {
+					$this->msg = esc_html__( 'Schema repositing only works in a Pantheon environment.', 'solr-for-wordpress-on-pantheon' );
+					break;
```
ataylorme: Typo? _repositing_ should be _reposting_

danielbachhuber: 8eaf334

PR #295, opened 2017-06-21
Handle `DOING_AJAX` and `is_admin()` conditionals separately

### LINKED ISSUE #284: Media Library search fails in "grid view" with solr-power enabled
Steps to reproduce: 

1. Add media, enable solr-power, create index
1. Search for a known string existing in the media library in the default "grid" view 

Expected outcome: 
- Results returned

Actual outcome: 
- "No media files found."

Workarounds: 
- Clicking the "list view" icon returns the expected results.

danielbachhuber:
`DOING_AJAX` is always `is_admin()`, so the prior implementation had
faulty logic.

If `DOING_AJAX=true`, we bail early unless the user has opted into using
Solr.

If `is_admin()` (and `DOING_AJAX` is false or unset), we bail early
unless the user has opted into using Solr.

Fixes #284

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -84,8 +84,8 @@ class SolrPower_WP_Query {
 	 * SolrPower_WP_Query instance initial setup method.
 	 */
 	function setup() {
-		// We don't want to do a Solr query if we're doing AJAX or in the admin area.
-		if ( ! is_admin() && defined( 'DOING_AJAX' ) && DOING_AJAX ) {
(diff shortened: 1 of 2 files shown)


danielbachhuber:
I've tested on http://dev-solr-power.pantheonsite.io/ and verified the facet widget still works with these changes:

facet

The key functional change is that searching the media library works as expected:

image

ataylorme:
@danielbachhuber I tested this as well, spinning up another multidev without the patch for comparison and saw the same results. Even adding custom fields, like `vehicle_cylinder` to the index and the facet widget is working as expected. Well done!

PR #299, opened 2017-07-10
Fix queries using orderby=>meta_value_num

### LINKED ISSUE #298: Issue with orderby = meta_value_num
Ordering the results by meta_value_num does not work for me when using Solr.

Example query:
$query_args = array( 'orderby'=>'meta_value_num', 'meta_key'=>'custom_number_field');

Everything is fine until function parse_orderby(...) in class-solrpower-wp-query.php line 272. On line 339 in the switch( $orderby ) case 'meta_value_num' it sets the $orderby_clause to "{$primary_meta_query['alias']}_i}. Unfortunately the "alias" is always = "wp_postmeta". So $orderby_clause then = "wp_postmeta_i" which is incorrect.

I changed line 340 from $orderby_clause = "{$primary_meta_query['alias']}_i"; to: $orderby_clause = "{$primary_meta_query['key']}_i"; and it works perfectly.

Is there anything I'm missing? Should line 340 use 'alias'? Unfortunately there were no Filters to update the $orderby_clause before it is used so I had to update the code directly.

Please consider changing line 340 to described as above.

Thanks
Dana

danielbachhuber:
Fixes #298

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,10 @@ To support searching by author name (e.g. where "Pantheon" would return posts au
 ```
 
 ## Changelog ##
+
+### 1.3.1 ###
+* Fixes Solr queries using `orderby=>meta_value_num` [[#299](https://github.com/pantheon-systems/solr-power/pull/299)].
+
 ### 1.3.0 ###
 * Add `.distignore` file for wp dist-archive solr-power
 * Make Solr in the admin opt-in only using the `solr_allow_admin` filter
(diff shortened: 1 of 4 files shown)


danielbachhuber:
> Please update the change log with the fix.

Updated.

> I would like to do this for all PRs going forward. It will make releases much quicker since I won't need to go back through merged items to update the change log on release

Sounds good.

PR #300, opened 2017-07-10
Fix use of 'fields'=>'ids', which should return post IDs

danielbachhuber:


danielbachhuber:
@ataylorme We have a bit of a problem with this, unfortunately.

Solr Power fills in posts from Solr on `the_posts` filter: https://github.com/WordPress/WordPress/blob/06fa4161aa74619239cf27017d124081c825684a/wp-includes/class-wp-query.php#L3007

When we return an array of IDs in `the_posts` filter, they get transformed into proper post objects: https://github.com/WordPress/WordPress/blob/06fa4161aa74619239cf27017d124081c825684a/wp-includes/class-wp-query.php#L3010-L3020

The workaround would be to use the `posts_pre_query` filter, added in WordPress 4.6. But, we need to increase the minimum version to 4.6: https://github.com/WordPress/WordPress/blob/06fa4161aa74619239cf27017d124081c825684a/wp-includes/class-wp-query.php#L2781

I can't think of an alternative fix at this point.

I noticed this issue while working on #299. There are lots of tests like this:

image

ataylorme:
@danielbachhuber per [Josh's comment in #284](https://github.com/pantheon-systems/solr-power/issues/284#issuecomment-308802691) I think bumping minimum version is okay. Please proceed.

cc @joshkoenig

danielbachhuber:
> per Josh's comment in #284 I think bumping minimum version is okay. Please proceed.

I confirmed today with Josh. I'll do so in a separate PR.

ataylorme:
@danielbachhuber what's the status of this work? I see the unit tests are failing. Wondering if we should try to get this into `1.4` which I'd like to release soon

danielbachhuber:
> what's the status of this work?

It's on my plate but I haven't started. I'll need to do a fair amount of refactoring to use `posts_pre_query`.

> Wondering if we should try to get this into `1.4` which I'd like to release soon

Is 1.5 to 2 weeks fine? I can work on the refactor early next.

ataylorme:
Saving it for `1.5` is fine. I'll be out on vacation until 9/5 starting next week but can review and cut a release when I return.

ataylorme:
@danielbachhuber this seems to be the last thing in the `1.5` milestone. Still plan on getting this in for that release?

danielbachhuber:
Closing in favor of #336

PR #303, opened 2017-07-12
Update Solarium to 3.8.1

### LINKED ISSUE #290: Update Solarium to 3.8.1
3.8.1 is the latest release https://github.com/solariumphp/solarium/releases/tag/3.8.1

Is there a reason we're pinned to 3.6.0?

danielbachhuber:
Fixes #290

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ To support searching by author name (e.g. where "Pantheon" would return posts au
 
 ### 1.4.0 ###
 * Bumps minimum supported version to WordPress 4.6.
+* Updates bundled Solarium library to 3.8.1.
 * Fixes Solr queries using `orderby=>meta_value_num` [[#299](https://github.com/pantheon-systems/solr-power/pull/299)].
 
 ### 1.3.0 ###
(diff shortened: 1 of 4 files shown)


danielbachhuber:
> Please update the change log as well

Done.

One thing to note with updating the changelog in each pull request is that we'll have merge conflicts between pull requests if there are multiple pull requests open. It might turn out easier to do them all at the end.

ataylorme:
> One thing to note with updating the changelog in each pull request is that we'll have merge conflicts between pull requests if there are multiple pull requests open. It might turn out easier to do them all at the end.

What about keeping an open PR for change log updates . My goal is to document changes as we go instead of recounting them retroactively and that should solve the problem.

danielbachhuber:
> What about keeping an open PR for change log updates .

Could work.

> My goal is to document changes as we go instead of recounting them retroactively and that should solve the problem.

Why isn't the list of issues assigned to the milestone sufficient for this?

https://github.com/pantheon-systems/solr-power/milestone/9?closed=1

I usually just copy the issue title into the changelog and make minor grammatical tweaks. It doesn't take more than a couple minutes.

ataylorme:
> Why isn't the list of issues assigned to the milestone sufficient for this?

That works as well we just need to be diligent about marking items in milestones, which hasn't happened in the past.

Issue #311, opened 2017-08-13
`Index custom fields` setting requires exact matches


jedkcanderson:
The `Index custom fields` setting allows users to add keys that should be added to the Solr index by naming the field keys directly.  This works great for simple fields added by ACF and other plugins that contribute additional fields to WordPress.

However, ACF allows for flexible content which contributes fields with keys that contain number, for example: ```programming_flex_content_2_programming_external_links_0_programming_external_link_label```

Given the flexibility of the data modeling, it is infeasible to add all of the possible combinations in the `Index custom fields` setting.

We'd like the ability to add a setting to allow for more flexible key matching.

danielbachhuber:
Replied in https://github.com/pantheon-systems/solr-power/pull/312#issuecomment-322246470

jedkcanderson:
Continuing the discussion started in #312 here.

ACF stores it's extra field data in the meta table using two keys for each piece of data you append.  One of the keys you don't want because it's just the glue that holds things together the value is usually something like `field_82938523`.

ACF supports a concept of a flexible content area where developers set up a list of available data types and authors can mix & match those fields as they see fit for any given page.  That's were the numbers I talk about above are added.  If you add two images, and a coupled video,and a text field you could get keys like:

```my_custom_post_type_content_0_image
my_custom_post_type_content_1_image
my_custom_post_type_content_2_video_group_content_0_video_url
my_custom_post_type_content_2_video_group_content_1_label```

Using a filter to register the field names of these fields is possible, but would require analyzing the data.  Is there a way to only run a filter if we're doing a `solr index`?  Otherwise I'd have to do a database crawl all the time which would slow everything else down.

ataylorme:
> Is there a way to only run a filter if we're doing a `solr index`? 

We have [an action for indexing being finished](https://github.com/pantheon-systems/solr-power/blob/ab406f0d918916f18d3238e9c9fa2ea170d92a69/includes/class-solrpower-sync.php#L652). 

I think we if add an action for indexing start you could hook into that action, do the analysis, add the filters and then remove the filters again in the `solr_power_index_all_finished` action.

What are your thoughts @danielbachhuber?

jedkcanderson:
Thinking about his more, maybe a different hook would make sense.  I'm going to end up querying the database for all of the documents that are being indexed.  That feels like we're duplicating work.

What about adding a filter like `solr_power_should_index` which takes a callback that accepts post data and a field name?  That way I can do my logic inline.

[...]

PR #312, opened 2017-08-13
Add extra setting to simplify the configuration

jedkcanderson:
Proposed feature addition allowing regexes to support complex field keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Solr Search for WordPress #
-**Contributors:** getpantheon, Outlandish Josh, 10up, collinsinternet, andrew.taylor  
+**Contributors:** getpantheon, Outlandish Josh, 10up, collinsinternet, andrew.taylor, jedkcanderson
 **Tags:** search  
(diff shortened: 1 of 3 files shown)


danielbachhuber:
Thanks for taking the time to submit a pull request @jedkcanderson.

Adding a regex pattern setting seems like it might be more complicated of a solution than we need to solve the use case identified in #311.

Instead, have you considered adding a filter to `solr_index_custom_fields` to programmatically insert your ACF fields?

jedkcanderson:
I figured regex was too big of a hammer.  😃 

Can you say more about your filter idea?

danielbachhuber:
> I figured regex was too big of a hammer.

More of a sledgehammer :)

> Can you say more about your filter idea?

I don't know ACF deeply, so I can't offer the specifics. With a function called `px_get_acf_field_names()` that returned all of the field names, the filter would look something like this:

```
add_filter( 'solr_index_custom_fields', function() {
     return px_get_acf_field_names();
});
```

PR #316, opened 2017-08-16
Rather than ignoring files with PHPCS skip them all together with find

ataylorme:


ataylorme:
@danielbachhuber opting in to only the directories we want to code sniff was easier than excluding others.

The minor downside here is that if we create new directories for Solr Power code they will need to be added to the opt in list for code sniffing.

danielbachhuber:
Hm. I think I'd prefer figuring out why PHPCS `<exclude-pattern>` stopped working in the first place.

ataylorme:
> I think I'd prefer figuring out why PHPCS <exclude-pattern> stopped working in the first place.

That would be nice to figure out but the build went much faster when `find` didn't have to go into `vendor`.

PR #319, opened 2017-08-25
Define checked files in PHPCS config so `phpcs` can easily be run

danielbachhuber:


.travis.yml
```diff
@@ -59,5 +59,5 @@ script:
     fi
   - |
     if [[ "$WP_TRAVISCI" == "phpcs" ]] ; then
-      phpcs --standard=phpcs.ruleset.xml --extensions=php .
```
ataylorme: Is `--standard=phpcs.ruleset.xml` implied?

danielbachhuber: No — but renaming the file to `phpcs.xml.dist` achieves the same.

Issue #326, opened 2017-09-26
Plugin page menu tab not working


sabdullah98:
Tabs on menu page not working after Ping is successful however if ping fails then tab starts working.

![solr1](https://user-images.githubusercontent.com/11044765/30858598-fa3a737a-a2dd-11e7-8725-da353a98ab24.png)
,
also even after succesful ping, as tabs are not workig, i tried WP-CLI and executed this command `wp solr index --allow-root
` all my product says `Failed to index '<product name>' (1961): Error 405 HTTP method POST is not supported by this URL`

danielbachhuber:
Hi @sabdullah98,

Thanks for the report.

> Tabs on menu page not working after Ping is successful however if ping fails then tab starts working.

Are there any JavaScript errors in your console that you can share?

>  i tried WP-CLI and executed this command `wp solr index --allow-root` all my product says `Failed to index '<product name>' (1961): Error 405 HTTP method POST` is not supported by this URL

Could you share more details about your Solr configuration?

sabdullah98:
`Tabs on menu page not working after Ping is successful however if ping fails then tab starts working.`
Here is the console screenshot of the page
![solr2](https://user-images.githubusercontent.com/11044765/30901984-5f671aec-a387-11e7-9643-e05b3060abc9.png)
Wordpress version 4.8.2 also tested on 4.6, same issue.

sabdullah98:
@danielbachhuber Any update on this Daniel?

danielbachhuber:
I can't reproduce this issue locally:

![solr](https://user-images.githubusercontent.com/36432/30964977-6c406e54-a408-11e7-8a5d-63394bd18e1b.gif)

Thinking through the console error:

* `replace` is used a couple times at the top of `admin.min.js`. Here's the uncompressed source: https://github.com/pantheon-systems/solr-power/blob/1.4.0/assets/js/src/admin.js#L3
* `replace` is contingent on `input[name="_wp_http_referer"]` being present on the page.
* "Indexing Stats by Post Type" is empty in the screenshot you've provided, which is unexpected.

Can you check your PHP error log to see if a fatal error is preventing the remaining part of the page from being rendered?

sabdullah98:
I am getting this PHP Fatal error
[solr-power error.txt](https://github.com/pantheon-systems/solr-power/files/1346283/solr-power.error.txt)

danielbachhuber:
[...]

PR #328, opened 2017-10-02
Catch exception when `$search->getData()` fails

danielbachhuber:
See #326


danielbachhuber:
> Please fix code sniffing errors in Travis

Fixed

PR #346, opened 2017-11-29
Support for indexing serialized arrays

mweimerskirch:
If the value is an array, recursively "implode" all values with a linebreak character.
Many plugins store data (e. g. repeatable text boxes) as serialized arrays in the post_meta table. This changes makes it easier to index them properly without indexing control characters or array index names.


danielbachhuber:
Hi @mweimerskirch,

Thanks for the pull request. Could you open an issue first so we can discuss before you spend too much time on the implementation? Including additional details on your use case, etc. would be great.

mweimerskirch:
@danielbachhuber I'm used to discussing these details right in the pull request, as comments can be added to the commit and I can later on easily "rebase" the code and replace the commit. 
But yes, I can create an issue.

cjcodes:
Two notes on this:

The schema won't like this--`*_s` does not support multiValued.

I would also recommend handling integer/float/decimal values, so you may want to move the checker up a bit.

And finally, `*_i`, `*_d`, and `*_f` don't support multiValued either, so would be worth submitting a patch against that as well.

(I'm handling these manually myself via the `solr_build_document` filter, but these suggestions would make it a bit easier.)

danielbachhuber:
@cjcodes Did you come up with a good solution for this that you'd want to submit a pull request for?

PR #362, opened 2018-06-01
Migrate CircleCI config to v2

danielbachhuber:
Based on https://github.com/pantheon-systems/pantheon-advanced-page-cache/pull/123
See https://github.com/pantheon-systems/pantheon-advanced-page-cache/issues/122

### DIFF
--- a/bin/install-solr.sh
+++ b/bin/install-solr.sh
@@ -45,7 +45,7 @@ download_and_run() {
     # Download from a Pantheon hosted file on CI
     if $CONTINUOUS_INTEGRATION
     then
-    	url="http://solr-power.pantheon.website/wp-content/uploads/apache-solr-3.6.2.tgz"
+    	url="https://dev-solr-power.pantheonsite.io/wp-content/uploads/apache-solr-3.6.2.tgz"
     else
     	url="http://archive.apache.org/dist/lucene/solr/3.6.2/apache-solr-3.6.2.tgz"
     fi
--- a/circle.yml
+++ b/circle.yml
@@ -1,50 +1,50 @@
-machine:
-  php:
-    # https://circleci.com/docs/environment#php
-    version: 5.6.22
-  environment:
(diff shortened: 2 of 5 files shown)


danielbachhuber:
@ataylorme Looks like `http://solr-power.pantheon.website/wp-content/uploads/apache-solr-3.6.2.tgz` disappeared (domain doesn't resolve). What's the best way to fix?

ataylorme:
@danielbachhuber you can use [https://dev-solr-power.pantheonsite.io/wp-content/uploads/apache-solr-3.6.2.tgz](https://dev-solr-power.pantheonsite.io/wp-content/uploads/apache-solr-3.6.2.tgz)

danielbachhuber:
> Please use the pantheon-public/build-tools-ci Docker image as it has Terminus pre-installed.

Does it have the other dependencies we need? Which PHP version is it?

ataylorme:
It is PHP 7.1 with all of our Terminus plugins and Composer. Not sure what else you need. You can see the Dockerfile [here](https://github.com/pantheon-systems/docker-build-tools-ci/blob/master/Dockerfile)

danielbachhuber:
Ok. WP SAML Auth requires PHP 5.6. I'd prefer to keep the `circle.yml` files consistent between plugin repos and don't see the benefit in using `pantheon-public/build-tools-ci` Docker image.

ataylorme:
The benefit is faster build since the Docker image has Terminus and it doesn't need to be installed on each CircleCI build.

danielbachhuber:
> The benefit is faster build since the Docker image has Terminus and it doesn't need to be installed on each CircleCI build.

Given the Composer checkout is cached, what percentage of total build time does installing Terminus take?

ataylorme:
The [last build](https://circleci.com/gh/pantheon-systems/solr-power/1189) took 4 minutes, which seems okay.

Moving the setup to its own file would still be nice but not necessary.

danielbachhuber:
👍 Thanks @ataylorme

PR #368, opened 2018-06-19
Added post__in and post__not_in support

ghamok:
Added in missing `post__in` and `post__not_in` WP_Query args using the approach described by @rdanamcd. However, I generalized the approach to account for other possible NOT queries and formatted code to meet project's code standards.

Cases That Should Now Work:
1. WP_Query with post_in arg (will work with array lengths of 1 or more)
2. WP_Query with post__not_in arg (will work with array lengths of 1 or more)

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -115,6 +115,9 @@ class SolrPower_WP_Query {
 
 		add_filter( 'posts_request', array( $this, 'posts_request' ), 10, 2 );
 
+		// Adjusts for NOT queries
+		add_filter( 'solr_select_query', array( $this, 'solr_select_query' ), 10, 1 ); 
+
 		// Nukes the FOUND_ROWS() database query.
 		add_filter( 'found_posts_query', array( $this, 'found_posts_query' ), 5, 2 );
 
@@ -358,6 +361,19 @@ class SolrPower_WP_Query {
 	}
 
 	/**
+	 * Adjusts the solr select query syntax.
+	 * 
+	 * @param string	$select		Select query
+	 * 
+	 * @return string
+	 */
+	function solr_select_query( $select ) { 
+		$select['query'] = str_replace( '(!', '!(', $select['query'] ); 
+		
+		return $select; 
+	}
+
+	/**
 	 * Overload the found posts query.
 	 *
 	 * @param string   $sql   Original SQL string.
@@ -447,13 +463,16 @@ class SolrPower_WP_Query {
 			'page_id',
 			'post_status',
 			'post_parent',
+			'post__in',
+			'post__not_in',
 			'name',
-
 		);
 		$convert = array(
-			'p'       => 'ID',
-			'page_id' => 'ID',
-			'name'    => 'post_name',
+			'p'       	=> 'ID',
+			'page_id' 	=> 'ID',
+			'post__in'	=> 'ID',
+			'post__not_in'	=> '!ID',
+			'name'    	=> 'post_name',
 		);
 		if ( ! $query->get( 's' ) && ! $query->get( 'solr_integrate' ) ) {
 			return '';
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -115,9 +115,6 @@ class SolrPower_WP_Query {
 
 		add_filter( 'posts_request', array( $this, 'posts_request' ), 10, 2 );
 
-		// Adjusts for NOT queries.
-		add_filter( 'solr_select_query', array( $this, 'solr_select_query' ), 10, 1 );
-
 		// Nukes the FOUND_ROWS() database query.
 		add_filter( 'found_posts_query', array( $this, 'found_posts_query' ), 5, 2 );
 
@@ -361,20 +358,6 @@ class SolrPower_WP_Query {
 	}
 
 	/**
-	 * Adjusts the solr select query syntax.
-	 *
-	 * @param string $select		Select query.
-	 *
-	 * @return string
-	 */
-	function solr_select_query( $select ) {
-		// Correcting whitelist NOT query conversion to proper solr syntax.
-		$select['query'] = str_replace( '(!ID', '!(ID', $select['query'] );
-
-		return $select;
-	}
-
-	/**
 	 * Overload the found posts query.
 	 *
 	 * @param string   $sql   Original SQL string.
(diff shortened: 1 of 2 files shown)


includes/class-solrpower-wp-query.php
```diff
@@ -357,6 +360,19 @@ private function parse_orderby( $orderby, $query ) {
...
 
+	/**
+	 * Adjusts the solr select query syntax.
+	 * 
+	 * @param string	$select		Select query
+	 * 
+	 * @return string
+	 */
+	function solr_select_query( $select ) { 
+		$select['query'] = str_replace( '(!', '!(', $select['query'] );
```
danielbachhuber: Can you include an inline comment documenting the type of transformation this is making?

danielbachhuber:
> I cannot guarantee an end of the day update, but it will be done long before the end of the week

Great, thanks!

ghamok:
HI @danielbachhuber , I added in the changes that you requested and fixed the phpcs issues.

includes/class-solrpower-wp-query.php
```diff
@@ -357,6 +360,19 @@ private function parse_orderby( $orderby, $query ) {
...
 
+	/**
+	 * Adjusts the solr select query syntax.
+	 *
+	 * @param string $select		Select query.
+	 *
+	 * @return string
+	 */
+	function solr_select_query( $select ) {
+		// Replace all occurrences of (! with !(.
```
danielbachhuber: This comment explains _what_ but it doesn't explain _why_. Can you expand your comment to explain why this change is necessary, and include an example of how it transforms a query?

Also, I'm curious as to why these occurrences _always_ need to be changed, instead of only when the query includes `post__in` or `post__not_in`.

Part of this is that I personally am still unclear on the nature of this change. I'd like to make sure I understand it fully before we merge.

tests/phpunit/wp_query/test-wp-query.php
```diff
@@ -178,4 +178,28 @@ public function test_wp_query_search_filter_post_type() {
 		$this->assertEquals( 'page', $query->get('post_type') );
 		$this->assertEquals( array( $page_id ), wp_list_pluck( $query->posts, 'ID' ) );
 	}
+
+	public function test_wp_query_by_post__in() {
+		$post_id = $this->__create_test_post();
```
danielbachhuber: Can you create a second `$post_id2` and make sure it's excluded from the result set?

tests/phpunit/wp_query/test-wp-query.php
```diff
@@ -178,4 +178,28 @@ public function test_wp_query_search_filter_post_type() {
...
+			'solr_integrate' => true,
+			'post__in'       => array( $post_id ),
+		);
+		$query = new WP_Query( $args );
+		$this->assertEquals( 1, $query->post_count );
+		$this->assertEquals( $post_id, $query->post->ID );
+	}
+
+	public function test_wp_query_by_post__not_in_arr() {
+		$post_id = $this->__create_test_post();
```
danielbachhuber: Can you create a second `$post_id2` and make sure it's included in the result set?

includes/class-solrpower-wp-query.php
ghamok: Even though I will correct the comment after the end of the work day, the transformation is mainly to correct invalid syntax. Without the filter, you would get a resulting select query of `(!ID:(8357 OR 8352))` which is invalid solr syntax. Correct syntax would be `!(ID:(8357 OR 8352))`.

In terms of always correcting the occurrence, I mainly did this to help future support of NOT related queries that are not necessarily ID-based. But if you would prefer to isolate the restriction to `!ID`, I can certainly make that change as well.

danielbachhuber: > But if you would prefer to isolate the restriction to `!ID`, I can certainly make that change as well.

Yes, this would be preferable, to limit the scope of the change and possibility of unintended side effects.

tests/phpunit/wp_query/test-wp-query.php
```diff
@@ -195,12 +195,14 @@ public function test_wp_query_by_post__in() {
...
 		SolrPower_Sync::get_instance()->load_all_posts( 0, 'post', 100, false );
 		$args  = array(
 			'solr_integrate' => true,
 			'post__not_in'   => array( $post_id ),
 		);
 		$query = new WP_Query( $args );
-		$this->assertEquals( 0, $query->post_count );
+		$this->assertEquals( 1, $query->post_count );
+		$this->assertEquals( $post_id2, $query->post->ID );
 	}
```
ghamok: `post__not_in` test currently fails where the `post_count` is zero instead of one. Currently trying to find out why, but my main concern is that a result of zero usually only occurred on my own solr instance (not this project's testing instance) when the !ID syntax was left uncorrected.

tests/phpunit/wp_query/test-wp-query.php
```diff
@@ -178,4 +178,31 @@ public function test_wp_query_search_filter_post_type() {
...
+		$post_id2 = $this->__create_test_post();
+		SolrPower_Sync::get_instance()->load_all_posts( 0, 'post', 100, false );
+		$args  = array(
+			'solr_integrate' => true,
+			'post__not_in'   => array( $post_id ),
+		);
+		$query = new WP_Query( $args );
+		$this->assertEquals( 1, $query->post_count );
+		$this->assertEquals( $post_id2, $query->post->ID );
+	}
```
ghamok: This test currently fails at the first assertion: results in a post_count of zero instead of 1. Currently trying to see if this an issue with the implementation or possibly a phpunit (as I am not that adept with phpunit) issue. My main concern is, that during my own testing (using the solr instance hosted on pantheon), getting no results only occurred when the new solr_select_query filter is not properly applied. 
Like I said, I will keep looking into this problem, however, do note that proper debugging on my end isn't 100% possible as setting up a local solr instance proved to be problematic.

ghamok:
Hi @danielbachhuber ,
I added the changes you request, however, I'm running into a rather odd issue. Currently, the post_in and post__not_in tests are nearly identical, yet only the post_in test runs to succession. The post__not_in test is failing at the first assertion where I'm testing the count of total posts retrieved. The job log is reporting that no posts at all are being retrieved which is concerning. 
I was wondering if you could shed some light on possible causes for this. The same code and the current version of this repo work fine on a Wordpress 4.9.6 pantheon multidev.

danielbachhuber:
@ghamok I was able to arrive at a slight different solution in 7d54529. We're good to go now. Thanks for your work on this!

PR #386, opened 2018-10-05
add ID param to solr_index_custom_fields filter

Ivan1234:


Ivan1234:
Sorry, that would not work

Issue #397, opened 2019-02-26
<h1> is much more helpful for debugging than <title> for batch indexing


krispy1298:
Proposing a change in [this line](https://github.com/pantheon-systems/solr-power/blob/master/includes/class-solrpower-batch-index.php#L241).

Here is output from a CLI index with an error:
```
Solr HTTP error: OK (400)\n<html><head><title>Apache Tomcat\/7.0.68 - Error report<\/title><style><!--H1 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:22px;} H2 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:16px;} H3 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:14px;} BODY {font-family:Tahoma,Arial,sans-serif;color:black;background-color:white;} B {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;} P {font-family:Tahoma,Arial,sans-serif;background:white;color:black;font-size:12px;}A {color : black;}A.name {color : black;}HR {color : #525D76;}--><\/style> <\/head><body><h1>HTTP Status 400 - ERROR: [doc=18807] unknown field 'valid_years'<\/h1><HR size="1" noshade="noshade"><p><b>type<\/b> Status report<\/p><p><b>message<\/b> <u>ERROR: [doc=18807] unknown field 'valid_years'<\/u><\/p><p><b>description<\/b> <u>The request sent by the client was syntactically incorrect.<\/u><\/p><HR size="1" noshade="noshade"><h3>Apache Tomcat\/7.0.68<\/h3><\/body><\/html>
```

`HTTP Status 400 - ERROR: [doc=18807] unknown field 'valid_years'` is much more helpful than `Apache Tomcat\/7.0.68 - Error report`.

danielbachhuber:
Sorry for the late reply here, @krispy1298. This sounds like a good suggestion. In fact, I'd suggest parsing the `<h1>` if it exists, otherwise falling back to `<title>`

danielbachhuber:
Fixed up in #407 407

PR #407, opened 2019-05-16
Issue #397: parse h1 tags in solr error response

pavellishin:
The <title> tags typically don't have any useful diagnostic information, but the <h1> tags do - so parse those out first!


danielbachhuber:
@pavellishin I wouldn't be bothered by mocking Solarium for this. You could abstract the parsing logic to a dedicated method if you wanted to write some simple unit tests for it.

pavellishin:
@danielbachhuber that's not a bad idea. I think I'd still want to add a test to ensure that SolrPower_Sync::post returning false invokes the error handler, but yanking that out would make it more easily extensible if there's more useful error messaging we can yank out of a response.

danielbachhuber:
@pavellishin On second thought, I think this PR looks fine as-is. Are you happy with it?

pavellishin:
@danielbachhuber yeah - I think your suggestion about pulling code out is good, but I could spend hours making little fixes forever :)

Issue #408, opened 2019-05-20
SolrPower_Api->get_server_info returns inaccurate server information


pavellishin:
https://github.com/pantheon-systems/solr-power/blob/master/includes/class-solrpower-api.php#L575-L584

`get_server_info()` assumes that the host name and port are the pantheon environmental variables, but it's possible to override the connection data via filters and constants.

(I'll try to get a PR out for this this week if I can.)

danielbachhuber:
> (I'll try to get a PR out for this this week if I can.)

That'd be great, thanks!

elliottpost:
I know there's a PR for this right now, but since it hasn't been approved yet, here's another option to solve this:

```php
    public function get_server_info() {
        $connection_options = apply_filters('s4wp_connection_options', [
            'endpoint' => [
                'localhost' => [
                    'host' => getenv( 'PANTHEON_INDEX_HOST' ),
                    'port' => getenv( 'PANTHEON_INDEX_PORT' ),
                    'path' => $this->compute_path()
                ]
            ]
        ]);

        $server_info = [
            'ping_status' => $this->ping_server(),
            'ip_address'  => $connection_options['endpoint']['localhost']['host'],
            'port'        => $connection_options['endpoint']['localhost']['port'],
            'path'        => $connection_options['endpoint']['localhost']['path'],
        ];

        return $server_info;

    }
```

jocastaneda:
I was tinkering with code trying to get a slightly better understanding of things as well and one possible way is also by changing the function to something like:

```php
public function get_server_info() {
	return array(
		'ping_status' => $this->ping_server(),
		'ip_address'  => $this->get_solr()->getEndpoint('localhost')->getHost(),
		'port'        => $this->get_solr()->getEndpoint('localhost')->getPort(),
		'path'        => $this->compute_path(),
	);
}
```

This would also mean changing the `get_solr()` function but that can be done like:

```php
$host = getenv( 'PANTHEON_INDEX_HOST' ) ? getenv( 'PANTHEON_INDEX_HOST' ) : null;
$port = getenv( 'PANTHEON_INDEX_PORT' ) ? getenv( 'PANTHEON_INDEX_PORT' ) : null;

$solarium_config      = array(
	'endpoint' => array(
		'localhost' => array(
			'host'   => apply_filters( 'solr_power_hostname', $host ),
			'port'   => apply_filters( 'solr_power_port', $port ),
			'scheme' => $this->get_default_scheme(),
[...]

PR #411, opened 2019-05-30
Read host and port information from solr instance, not Pantheon env variables

pavellishin:
See issue #408 

- [ ] Need tests

### DIFF
--- a/tests/phpunit/test-solrpower-api.php
+++ b/tests/phpunit/test-solrpower-api.php
@@ -1,0 +1,40 @@
+<?php
+
+class SolrPowerAPITest extends SolrTestBase {
+
+	function setUp() {
+		parent::setUp();
+	}
+
+	function tearDown() {
+		parent::tearDown();
+	}
+
+
+	function test_server_info_uses_actual_config() {
+		$env_host      = $ENV['PANTHEON_INDEX_HOST'];
+		$env_port      = $ENV['PANTHEON_INDEX_HOST'];
+		$override_host = $env_host . 'OVERRIDE';
+		$override_port = $env_host . 'OVERRIDE';
+
+		$override_connection = function() {
+			return array(
+				'endpoint' => array(
+					'localhost' => array(
+						'host' => $override_host,
+						'port' => $override_port,
+						'path' => '',
+					)
+				),
+			);
+		};
+		add_filter( 's4wp_connection_options', $override_connection );
+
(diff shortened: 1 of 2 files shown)


pavellishin:
@danielbachhuber I'm having some issues with my tests - they don't see `PANTHEON_INDEX_HOST` set in the env. (You can see my past few commits trying to fiddle around with this.) Shouldn't `bootstrap.php` be taking care of that?

tests/phpunit/test-solrpower-api.php
```diff
@@ -0,0 +1,40 @@
...
+		parent::setUp();
+	}
+
+	function tearDown() {
+		parent::tearDown();
+	}
+
+
+	function test_server_info_uses_actual_config() {
+		$env_host      = array_key_exists('PANTHEON_INDEX_HOST', $_ENV) ? $_ENV['PANTHEON_INDEX_HOST'] : '';
```
ataylorme: > I'm having some issues with my tests - they don't see PANTHEON_INDEX_HOST set in the env. (You can see my past few commits trying to fiddle around with this.) Shouldn't bootstrap.php be taking care of that?

@pavellishin try using `getenv` instead of the `$_ENV` global

pavellishin: Oh, duh, I thought the previous code accessed it directly :P

tests/phpunit/test-solrpower-api.php
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+
+	function test_server_info_uses_actual_config() {
+		$env_host      = getenv('PANTHEON_INDEX_HOST');
+		$env_port      = getenv('PANTHEON_INDEX_PORT');
+		$override_host = $env_host . 'OVERRIDE';
+		$override_port = $env_host . 'OVERRIDE';
+
+		$override_connection = function() {
```
ataylorme: @pavellishin I don't think `$override_host` is defined in the context of this function. Maybe pass it in as a parameter?

pavellishin: oh, dangit. I forget how modern PHP works wrt scoping.

pavellishin:
Sorry I abandoned this for a few weeks, busy times at work and home :/

pavellishin:
Argh, I gotta set up phpunit and a test wordpress DB locally.

PR #444, opened 2020-03-26
Tab in the readme.txt code block

barryhughes:
Minor formatting change: tabs in one of the code blocks instead of using triple backticks, so that things render nicely [over here](https://wordpress.org/plugins/solr-power/#installation) :-)


danielbachhuber:
Thanks @barryhughes !

PR #446, opened 2020-04-23
added support for changing post excerpt to highlighted snippet 

jimkeller:
Highlighting is enabled in the solr query but it didn't look like it was ever making it back to the post object for use in the search result. This patch brings the highlighted snippet into the $post->excerpt.


danielbachhuber:
Thanks @jimkeller!

I'll go ahead and tag a new release with this.

PR #448, opened 2020-04-28
Use posts_pre_query hook to support fields queries

### LINKED ISSUE #336: Refactor query filtering to use 'posts_pre_query' filter introduced in WP 4.6
Originally https://github.com/pantheon-systems/solr-power/pull/300#issuecomment-314176820

Notably, `'fields'=>'ids'` won't work for `WP_Query` until this is complete because `WP_Query` runs `$this->posts = array_map( 'get_post', $this->posts );` against the post list.

mattleff:
Fixes #336. Add support for using Solr results in WP-REST API.

I'm not a WP expert so I might not have implemented this correctly. Please let me know any changes that need to be made. There are a few tests that fail locally when run with the whole suite, but pass when run individually.

### DIFF
--- a/tests/phpunit/wp_query/test-tax-query.php
+++ b/tests/phpunit/wp_query/test-tax-query.php
@@ -58,8 +58,8 @@ class SolrTaxQueryTest extends SolrTestBase {
 	function test_wp_query_by_tax_id() {
 		$t  = self::factory()->term->create( array(
 			'taxonomy' => 'genre',
-			'slug'     => 'horror',
-			'name'     => 'Horror',
+			'slug'     => 'biography',
+			'name'     => 'Biography',
 		) );
 		$this->__create_test_post();
 
@@ -174,7 +174,7 @@ class SolrTaxQueryTest extends SolrTestBase {
 			),
 		) );
 
-		$this->assertEquals( array( $p1 ), wp_list_pluck( $q->posts, 'ID' ) );
+		$this->assertEquals( array( $p1 ), $q->posts );
 	}
 
 	public function test_tax_query_single_query_single_term_field_name() {
@@ -202,7 +202,7 @@ class SolrTaxQueryTest extends SolrTestBase {
 			),
 		) );
 
-		$this->assertEquals( array( $p1 ), wp_list_pluck( $q->posts, 'ID' ) );
+		$this->assertEquals( array( $p1 ), $q->posts );
 		$facets = SolrPower_WP_Query::get_instance()->facets;
 		$this->assertEquals( 1, $facets['categories']->getValues()['Foo^^'] );
 		$this->assertCount( 0, $facets['tags']->getValues() );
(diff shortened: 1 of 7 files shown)


tests/phpunit/wp_query/test-tax-query.php
```diff
@@ -58,8 +58,8 @@ function test_wp_query_by_tax_slug() {
 	function test_wp_query_by_tax_id() {
 		$t  = self::factory()->term->create( array(
 			'taxonomy' => 'genre',
-			'slug'     => 'horror',
-			'name'     => 'Horror',
+			'slug'     => 'biography',
```
mattleff, line 61: This was the only test change of substance. The original test fails with a duplicate term error. The duplicate term seems to be coming from: 

https://github.com/pantheon-systems/solr-power/blob/458cd5bd889e38baf57cab6da9535b9ec4417a8d/tests/phpunit/class-solr-test-base.php#L87-L89

I'm not sure how it worked before...

danielbachhuber:
@mattleff Thanks for the pull request! I'll do a further review when I have a moment.

danielbachhuber:
> I'm not a WP expert so I might not have implemented this correctly. Please let me know any changes that need to be made.

I think it looks pretty good, actually. I'm surprised at how simple of a change it was. It'd be great to have some production smoke-testing before we tag a release though. Do you have a reasonably large site you could try this out on?

> There are a few tests that fail locally when run with the whole suite, but pass when run individually.

As it turns out, I think this is an issue with our use of `spl_object_hash()`: 56ef4ecfa5b556b548979a39f46c328f8b77c204

Namely, `spl_object_hash()` [doesn't do what you'd expect it to do](https://www.php.net/manual/en/function.spl-object-hash.php#76220):

> Note that the contents (properties) of the object are NOT hashed by the function, merely its internal handle and handler table pointer. This is sufficient to guarantee that any two objects simultaneously co-residing in memory will have different hashes. Uniqueness is not guaranteed between objects that did not reside in memory simultaneously, for example:
>
> var_dump(spl_object_hash(new stdClass()), spl_object_hash(new stdClass()));
> 
> Running this alone will usually generate the same hashes, since PHP reuses the internal handle for the first stdClass after it has been dereferenced and destroyed when it creates the second stdClass.

mattleff:
@danielbachhuber Thanks for reviewing this and getting the tests passing! I ran into this issue on a site that's in development. Let me check with my dev lead and see if we have a production site we could try this PR on.

mattleff:
@danielbachhuber I checked with my dev lead and we don't have another production site using this plugin atm. I did confirm that the latest code solves the issue we encountered on our development site (Gutenberg link autocompletes MIA), and I didn't notice any regressions on our search views. Let me know if there's anything else I can do to help test this out.

danielbachhuber:
@mattleff Cool. I think I feel fairly confident about it. I'm going to give it another look later this week and then tag if I feel good.

danielbachhuber:
@mattleff Just tagged v2.2.0. Thanks for your contribution!

PR #463, opened 2020-07-23
Add post_date_iso to schema

MikeNGarrett:
Using the TrieDateField field type for this date gives developers the ability to run boosts with date calculations. 
For example: `recip(abs(ms(NOW/HOUR,post_date_iso)),3.16e-11,1,1)`

The existing date fields cause a failurl when attempting to run functions like `ms()`.

### DIFF
--- a/includes/class-solrpower-sync.php
+++ b/includes/class-solrpower-sync.php
@@ -297,6 +297,7 @@ class SolrPower_Sync {
 			$doc->setField( 'hour_i', gmdate( 'H', $post_time ) );
 			$doc->setField( 'minute_i', gmdate( 'i', $post_time ) );
 			$doc->setField( 'second_i', gmdate( 's', $post_time ) );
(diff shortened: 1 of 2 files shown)


danielbachhuber:
Thanks @MikeNGarrett !

Will changing the schema (and adding a field sent in the sync function) cause any errors for existing Solr users?

MikeNGarrett:
@danielbachhuber Yes, indexing will fail. I see a good path to support this on Pantheon. How do you suggest handling it for users who have implemented their own Solr instance?

danielbachhuber:
@MikeNGarrett To be honest, we're not really actively adding new features to Solr Power at the moment. The best way to add this might be to simply document how it's possible in the README. It looks like you could filter this attribute into the Solr document with the `solr_build_document` filter. Want to submit a pull request for that instead?

MikeNGarrett:
Got it. Will do.

PR #466, opened 2020-07-27
464 publish date

### LINKED ISSUE #464: Support TrieDateField for publish date
Current schema does not support a date field we can use to run date functions against. Adding TrieDateField support in the schema would support publish date boosting and other date-based calculations. 
#463

MikeNGarrett:
Fixes #464 by changing `readme.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # Solr Search for WordPress #
-**Contributors:** getpantheon, Outlandish Josh, 10up, collinsinternet, andrew.taylor, danielbachhuber, mattleff  
-**Tags:** search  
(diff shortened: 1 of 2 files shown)


danielbachhuber:
@MikeNGarrett It looks like there's a lot more in this diff than I'd expect. Also, the README.md is built from the readme.txt; sorry I didn't mention that before.

Can you add the change to readme.txt, and then build README.md?

MikeNGarrett:
@danielbachhuber Switched.

danielbachhuber:
@MikeNGarrett Looks great, thanks! I appreciate the 'Common issues' section too.

PR #473, opened 2020-11-25
Move build matrix to CircleCI

danielbachhuber:


danielbachhuber:
@kshaner Any ideas on what I might need to do in order to interface with your Solr docker image correctly?

kshaner:
@danielbachhuber I assume you would just have to target the container on docker hub and mount or copy the solr power schema file into the container at /opt/apache-solr-3.6.2/example/solr/conf/schema.xml. My dockerfile is fairly straightforward and was based on the bash install script in the repo. I'm not a CircleCI user myself, so I'm not 100% sure of their capabilities.

When I use this I have a docker-compose file at the root of the site that looks like this:

```yml
version: '3'

services:
    solr:
        image: kshaner/solr:latest
        ports:
            - '8988:8983'
        volumes:
            - ./wp-content/uploads/solr-for-wordpress-on-pantheon/schema.xml:/opt/apache-solr-3.6.2/example/solr/conf/schema.xml
            - solr-data:/opt/apache-solr-3.6.2/example/solr
volumes:
    solr-data:
```

If not using a custom schema, I just mount the schema file directly from the solr-power directory and the non standard port is only if using multiple instances, it defaults to 8983.

danielbachhuber:
> I assume you would just have to target the container on docker hub and mount or copy the solr power schema file into the container at /opt/apache-solr-3.6.2/example/solr/conf/schema.xml.

Ah, yeah that is the nut of the problem:

image

It looks like CircleCI doesn't support the same volume mounting approach you take: https://support.circleci.com/hc/en-us/articles/360007324514-How-can-I-use-Docker-volume-mounting-on-CircleCI-

Hm...

kshaner:
Looks like you could also use the `docker cp` command as shown here: https://circleci.com/docs/2.0/building-docker-images/#mounting-folders. The schema file in this instance is very similar to a config file.

danielbachhuber:
@kshaner Thanks for the tip. I tried that (see above) but it looks like I'm going to run into the separation of environments issue: https://circleci.com/docs/2.0/building-docker-images/#separation-of-environments

I suppose I'll just need to build my own container.

danielbachhuber:
@kshaner Thanks again for your help! I managed to get it working 💪

kshaner:
Nice! that was going to be next suggestion. I tend to use a lot of custom schemas so I wouldn't want to change my build to a hard coded schema file.

PR #478, opened 2020-12-03
Readme files: Fixing subcommand options

yuri-pochepko:
Bring Readme files to a current state for a index subcommand.

We also need to change it at https://wordpress.org/plugins/solr-power/


danielbachhuber (approved):
👍 Great catch!

PR #500, opened 2021-07-27
Remove style and script tag content

PatelUtkarsh:
`strip_tags` preserve content from style and script tags, Some of the Gutenberg blocks like to have inline script / style which appears in search results.

### DIFF
--- a/includes/class-solrpower-sync.php
+++ b/includes/class-solrpower-sync.php
@@ -270,7 +270,7 @@ class SolrPower_Sync {
 			}
 			$doc->setField( 'post_name', $post_info->post_name );
 			$doc->setField( 'post_title', $post_info->post_title );
-			$doc->setField( 'post_content', strip_tags( $post_info->post_content ) );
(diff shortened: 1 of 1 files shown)


danielbachhuber:
Hey @PatelUtkarsh !

> `strip_tags` preserve content from style and script tags, Some of the Gutenberg blocks like to have inline script / style which appears in search results.

I'm not 100% I understand. Can you share an example of what you're trying to solve for?

PatelUtkarsh:
Hello @danielbachhuber :) 

We are having a style tag as the starting of the `post_content` which results in an output of style as the post doesn't have `post_excerpt` details filled in causing style content to show up in the result when using with solr.

Example difference: 

> wp_strip_all_tags('<style>.bold{ font-weight: 700; }</style>')

Outputs to '' (empty string)

> strip_tags('<style>.bold{ font-weight: 700; }</style>');

Outputs to `.bold{ font-weight: 700; }`

danielbachhuber (approved):
👍 Makes sense, thanks for the explanation! I'll have v2.2.5 tagged shortly.

Issue #512, opened 2022-02-22
PHP 8 Deprecation Notices: Required parameter $section follows optional parameter $page


kshaner:
With WP_DEBUG on there are deprecation notices on front and back end:

```
PHP Deprecated:  Required parameter $section follows optional parameter $page in /var/www/wp-content/plugins/solr-power/includes/class-solrpower-options.php on line 589
```
Since it's a private method, and all usage seems to explicitly specify the $page, $section, and $type arguments, it would appear to be safe to specify a default value for $section and $page which gets rid of these deprecation notices.

danielbachhuber:
🔨

PR #513, opened 2022-02-22
Fix PHP 8 deprecations in class-solrpower-options.php

kshaner:
Fixes deprecation notices mentioned in #512


danielbachhuber (approved):
Thanks @kshaner !

PR #521, opened 2022-03-25
Found posts don't need to be mapped for posts_pre_query

### LINKED ISSUE #520: Issue with page search because of hierarchy and id=>parent fields
Our page search has been broken and all arrows point to the`posts_pre_query` filter in the `SolrPower_WP_Query` class.

The page search is different because it is a hierarchical post type. And when a hierarchical post type is searched, it uses `id=>parent` as the return fields parameter.

In the plugin's `posts_pre_query` filter, it converts "found posts" to just ID's before returning that info to core. So then core goes to map out the `id=>parent` and it breaks because all it has is an ID instead of a post object.

I will do a PR with the simple fix I made to correct this issue. So far I haven't discovered if this change creates any other conflicts but I will let you know if I do. I, of course, would love to know if you know of any issues with this change.

Thanks!

bamadesigner:
Found posts don't need to be mapped for `posts_pre_query` filter. This mapping is taken care of by the core query.

Fixes #520.

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -463,14 +463,7 @@ class SolrPower_WP_Query {
 			return null;
 		}
 
-		$new_posts = $this->found_posts[ $query->solr_query_id ];
-
-		return array_map(
-			function ( $post ) {
-				return $post->ID;
-			},
-			$new_posts
-		);
+		return $this->found_posts[ $query->solr_query_id ];
 	}
 
 	/**


danielbachhuber:
Thanks @bamadesigner!

I'll review this at depth when I have a moment in the next week or two. It seems like a substantial change, so I want to make sure I understand it fully. If you can, it would be great to have a couple of unit tests documenting the nature of the change.

danielbachhuber (approved):
Looks fine after a second look!

1. Core's `array_map( 'get_post' )` has been around for a long time: https://github.com/WordPress/wordpress-develop/blame/66ac1f40ed3d1191f7513790c34ccbe39a203fb2/src/wp-includes/query.php#L3671-L3674
2. `fields=>ids` expects an array of post ids, so Solr Power's implementation has been incorrect for a while: https://github.com/WordPress/wordpress-develop/blame/66ac1f40ed3d1191f7513790c34ccbe39a203fb2/src/wp-includes/query.php#L3593-L3598

There's some relevant past conversation in #336 and #448

danielbachhuber:
As it turns out, the test suite didn't run on this PR and there were a whole host of explosions when it did 🙃 

I've fixed things up with 2f312cf466e08ef3ba35f24fcd1ea16a05fd014d (#522 )

Thanks again for the PR!

bamadesigner:
Thanks @danielbachhuber!

PR #522, opened 2022-03-29
Adapt posts_pre_query() return values based on 'fields' argument

danielbachhuber:
From #521

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 **Requires at least:** 4.6  
 **Requires PHP:** 7.1  
 **Tested up to:** 5.9  
-**Stable tag:** 2.3.0  
+**Stable tag:** 2.3.1  
 **License:** GPLv2 or later  
 **License URI:** http://www.gnu.org/licenses/gpl-2.0.html  
 
@@ -217,6 +217,9 @@ Add the following to your `functions.php` file.
 
 
 ## Changelog ##
+
+### 2.3.1 (March 29, 2022) ###
+* Adapts `posts_pre_query()` return values based on 'fields' argument [[#522](https://github.com/pantheon-systems/solr-power/pull/522)].
 
 ### 2.3.0 (March 29, 2022) ###
 * Removes incorrect use of `array_map( 'get_post' )` in `posts_pre_query` [[#521](https://github.com/pantheon-systems/solr-power/pull/521)].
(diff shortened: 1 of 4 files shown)


bamadesigner:
Hey @danielbachhuber! I saw the update! Thank you!

But it still doesn't work for me.

So the `page` post type is hierarchical and when the `page` posts are queried for the "Edit pages" admin page (`/wp-admin/edit.php?post_type=page`), core sets the `fields` query parameter to `id=>parent`. 

As seen here: https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/post.php#L1232

So our issue comes when searching pages in the admin.

In this plugin's `posts_pre_query` filter (that you edited), if `id=>parent` is set, the filter maps the posts array of WP_Post objects to an ID (key) >  parent ID (value) array and returns that array to define `$this->posts` in the core query.

As seen here: https://github.com/pantheon-systems/solr-power/blob/f70e2b96889db8933d126d4156589410be8aa845/includes/class-solrpower-wp-query.php#L478-L480

So in the core query, when `id=>parent` is set, it loops through the posts array and needs access to each post's `ID` and `post_parent` property. Which is not available because this plugin's `posts_pre_query` removed them in its mapping.

Core code is here: https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L3057-L3075

So this breaks our page search query and I'm left wondering if there is something in our environment causing this issue, but in examining the code in relation to this plugin, along with a bit of localized debugging, it seems fairly obvious to me that the plugin is the issue.

I would very much welcome and appreciate your insight/thoughts.

danielbachhuber:
> So in the core query, when `id=>parent` is set, it loops through the posts array and needs access to each post's `ID` and `post_parent` property. Which is not available because this plugin's `posts_pre_query` removed them in its mapping.

Dang. That does seem incorrect. I was able to reproduce the failure with ef9a574ab9e20aeb98de6a23efd3be0d5836d124

I've added a fix with 9c9f113a7523c58f141434905af8cca212891e95 and tagged v2.3.2.

Can you see if this fixes your issue? If it doesn't, can you share more detail about the buggy behavior you're experiencing, and the debugging you've done?

bamadesigner:
Your fix works! Thank you!

danielbachhuber:
😅 Awesome! 

Thanks for your help tracking it down 😊

PR #554, opened 2022-11-18
[CMS-1033] Update CONTRIBUTING.md

jspellman814:


CONTRIBUTING.md
```diff
@@ -19,10 +23,12 @@ Note that dependencies are installed via Composer and the `vendor` directory is
...
-5. Wait for CI to build and push a new tag.
-6. Confirm that the necessary assets are present in the newly created tag, and test on a WP install if desired.
-7. Publish a new release using the latest tag. Publishing a release will kick off `wordpress-plugin-deploy.yml` and release the plugin to wp.org. If you do not want a tag to be publised to wp.org, do not publish a release from it.
\ No newline at end of file
+1. Update the Changelog with the latest changes.
+1. Create a PR against the `main` branch.
+1. After all tests pass and you have received approval from a CODEOWNER (including resolving any merge conflicts), merge the PR into `main`.
+1. [Check the _Build and Tag_ action](https://github.com/pantheon-systems/solr-power/actions/workflows/build-tag.yml): a new tag named with the version number should've been created. It should contain all the built assets.
+1. Create a [new release](https://github.com/pantheon-systems/solr-power/releases/new), naming the release with the new version number, and targeting the tag created in the previous step. Paste the release changelog from `CHANGELOG.md` into the body of the release and include a link to the closed issues if applicable.
```
jazzsequence, line 32: Can we automate the Release step, too?

jazzsequence (approved):
LGTM if it all works as advertised! :shipit:

CONTRIBUTING.md
greg-1-anderson: I wrote a [simple release script](https://github.com/consolidation/self-update/blob/main/scripts/release) a while back that might be useful.

jspellman814: There are also quite a few GH Actions available to automate Releases (this one, for example, https://github.com/marketplace/actions/automatic-releases), we'd just need to determine _when_ we'd like a release to be automated (e.g. every time a tag is created). I like the current workflow proposal because it gives us a little more control on when a release it published, but open to suggestions!

Issue #566, opened 2023-01-05
Divide by zero warning when posting a query


threeiem:
There is an assumption that there is more than one post to calculate posts_per_page in `class-solrpower-wp-query.php`.  This results in a warning.

```txt
PHP Warning:  Division by zero in /code/wp/wp-content/plugins/solr-power/includes/class-solrpower-wp-query.php on line 246
```

scottbuscemi:
This has been fixed in today's 2.5.1 release, thank you for reporting!

PR #567, opened 2023-01-06
[CMSP-42] Fixed issue with empty/zero value for posts_per_page.

threeiem:
* Set posts_per_page to default to 1 if not set
* Some whitespace changes

Fixes: #566


namespacebrian:
Tracking internally as CMS-1185

PR #568, opened 2023-01-09
[BUGS-5735] Include schema.xml in release distribution

jspellman814:


jazzsequence (approved):
LGTM, any idea why the Behat tests are failing at a password prompt? Would using GH Actions be any improvement?

jspellman814:
@jazzsequence it's failing because of key removal/rotation due to New Relic incident, but I'm working to fix. I'm assuming we'll run into this a lot of the next month or so 😅

jazzsequence:
@jspellman814 I wasn't sure if it was tied to that. If that's the case I'm less worried, but I also wonder if we should migrate away from Circle anyway as a rule.

jspellman814:
I'm totally down for that 🚀 We have a ticket to track that work (1131) that'd I could get started on (would work on that implementation in a separate PR, though).

PR #571, opened 2023-01-19
[CMSP-93] Migrate to GitHub actions

jspellman814:


.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,148 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/pantheon-public/build-tools-ci:8.x-php7.4
+      options: --user root
+    name: Lint
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
```
kporras07, line 19: Use v3 if possible as v2 is currently being deprecated

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,148 @@
...
+
+    - name: Cache dependencies
+      uses: actions/cache@v2
```
kporras07, line 22: Use v3 if possible as v2 is currently being deprecated

.github/workflows/lint-test.yml
jspellman814: Will do, thanks!

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,148 @@
...
+      run: composer phpcs
+
+  test-phpunit:
+    needs: lint
+    runs-on: ubuntu-latest
+    services:
+      mariadb:
+        image: mariadb:10.3
+      solr:
+        image: solr:8.3.1
```
kporras07, line 41: Do you really need this service? I see that later you're installing Solr in the same container image

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,148 @@
...
+        composer phpunit
+        rm -rf $WP_TESTS_DIR $WP_CORE_DIR
+        bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1 nightly true
+        WP_MULTISITE=1 composer phpunit
+
+  test-behat:
+    needs: test-phpunit
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/pantheon-public/build-tools-ci:6.x
```
kporras07, line 93: Any chance you take this opportunity to update to 8.x docker image? 8.x-php7.4 or 8.x-php8.0 as needed

kporras07 (approved):
LGTM only a couple of comments/questions

.github/workflows/lint-test.yml
jspellman814: I've removed this and [unit tests pass](https://github.com/pantheon-systems/solr-power/actions/runs/4056331799/jobs/6980727441) without issue.

bin/behat-prepare.sh
```diff
@@ -24,6 +24,10 @@ fi
 
 set -ex
 
+# Set StrictHostKeyChecking to no
+echo "StrictHostKeyChecking no" >> /etc/ssh/ssh_config
+cat /etc/ssh/ssh_config
```
pwtyler, line 29:
```suggestion
```
Drop the cat if it was only for debugging at the time

bin/install-wp-tests.sh
```diff
@@ -5,9 +5,9 @@ if [ $# -lt 3 ]; then
...
 
-DB_NAME=$1
-DB_USER=$2
-DB_PASS=$3
+DB_NAME=solr-power
+DB_USER=root
+DB_PASS=root
 DB_HOST=${4-localhost}
 WP_VERSION=${5-latest}
 SKIP_DB_CREATE=${6-false}
```
pwtyler, line 13: You are ignoring the first three args here and still passing six arguments to the script above. if these are not going to be provided args, these (and the invocation) should be adjusted.

bin/behat-prepare.sh
```diff
@@ -43,15 +47,28 @@ BASH_DIR="$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ###
 terminus connection:set $SITE_ENV git
 rm -rf $PREPARE_DIR
-git clone -b $TERMINUS_ENV $PANTHEON_GIT_URL $PREPARE_DIR
+git clone -b $TERMINUS_ENV $PANTHEON_GIT_URL $PREPARE_DIR -v
```
pwtyler, line 50: why verbose?

bin/behat-prepare.sh
jspellman814: Addressed in #572

bin/behat-prepare.sh
jspellman814: Addressed in #572

bin/install-wp-tests.sh
jspellman814: Addressed in #572

PR #575, opened 2023-03-20
[CMSP-73] PHP 8.2 compatibility

jspellman814:
This will require the minimum supported PHP version to be bumped from 7.1 to 7.4.


jasonyarrington:
Any progress on this?  I am looking to use this plugin on a site on PHP 8.

PR #587, opened 2023-06-07
Remove not necessary publish_post action.

kporras07:


includes/class-solrpower-sync.php
```diff
@@ -41,7 +41,6 @@ public static function get_instance() {
 	 * Instantiate the object.
 	 */
 	function __construct() {
-		add_action( 'publish_post', array( $this, 'handle_modified' ) );
```
kporras07, line 44: As per https://github.com/WordPress/WordPress/blob/f08039702ecdfeaba850c91bc0e3008272104509/wp-includes/post.php#L4892 both actions will run one after the other one in case of a save_post.

includes/class-solrpower-sync.php
```diff
@@ -41,7 +41,6 @@ public static function get_instance() {
...
 		add_action( 'publish_page', array( $this, 'handle_modified' ) );
```
ryanshoover, line 44:
```suggestion
```
Similar to publish_post, we don't need publish_page either, right?

kporras07: There is no save_page here. Maybe we should swap it with `save_page`? Not sure but IIUC save_post won't cover publish_page

ryanshoover: `do_action( 'publish_page' )` gets called by [wp_transition_post_status](https://github.com/WordPress/WordPress/blob/f08039702ecdfeaba850c91bc0e3008272104509/wp-includes/post.php#L5418) dynamically, which is called by wp_publish_post. 

`do_action( 'save_post' )` gets called further down in [wp_publish_post](https://github.com/WordPress/WordPress/blob/f08039702ecdfeaba850c91bc0e3008272104509/wp-includes/post.php#L4904). It also gets called by [wp_insert_post](https://github.com/WordPress/WordPress/blob/f08039702ecdfeaba850c91bc0e3008272104509/wp-includes/post.php#L4730)

The only reason we'd need both publish_page and save_post is if somebody was manually calling wp_transition_post_status.

PR #588, opened 2023-06-29
[CMSP-486] Add is_solr_query filter

sboisvert:
Right now it's impossible to turn off Solr for specific types of searches. 

Another considered solution was to allow for $query->get( 'solr_integrate' ); to be set to false and have that overwrite the result of $query->is_search() but the filter seems like it would give more flexibility and be clearer in it's goal.

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -1245,7 +1245,16 @@ class SolrPower_WP_Query {
 	 * @return boolean
 	 */
 	private function is_solr_query( $query ) {
-		return $query->is_search() || $query->get( 'solr_integrate' );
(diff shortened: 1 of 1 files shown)


rwagner00:
I created a ticket to look at this internally, thank you for the PR!

jazzsequence:
I love this. @rwagner00 can you update this with the Jira # and/or update the title.

rwagner00:
@jazzsequence  The ticket is: CMSP-486

PatelUtkarsh:
I'd also love to have some control using this filter to exclude specific rest api requests.

jazzsequence (approved):

pwtyler (approved):
Not sure why Polyfills is unhappy, pushed to #638 and got a green test. I'll followup to address changelog + doc + minor version bump.

PR #590, opened 2023-06-30
Add WP.org validator

jazzsequence:


CdrMarks (approved):

.github/workflows/wporg-validator.yml
```diff
@@ -0,0 +1,13 @@
...
+name: WP.org Validator
+on: [push]
+jobs:
+  wporg-validation:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: WP.org Validator
+        uses: pantheon-systems/action-wporg-validator@1.0.0
```
pwtyler, line 11: Can we use next significant release operators in github actions to always use the latest 1.x version, or do we need to pin to a specific version of the validator?

jazzsequence: need to pin to specific vesion, at least for now. I need to look at how to do `@latest` or whatever else.

PR #591, opened 2023-07-05
[BUGS-6482] Disable Auto-commit by default

pwtyler:
The default behavior is now to allow cron to push solr commits regularly instead of autocommit. See #559 and #549 for details when this option was initially added last year— now we're making it the default behavior.

Users can still enable auto-commit by explicitly setting `SOLRPOWER_DISABLE_AUTOCOMMIT` to `false`.

PR #592, opened 2023-07-10
Revert "Add WP.org validator (#590)"

pwtyler:
This reverts commit ef02320426995d65b4f7ae098d4479a999152100.

Something in #590 caused the widget to disappear on the test site when 2.5.0-dev was installed. Discovered while testing #591. 

Note, we will need to make sure we update or write behat tests that cover this better when re-implemented.


CdrMarks (approved):
Approving the revert.  Haven't tested.

PR #599, opened 2023-07-25
[CMSP-42] Fixed issue with empty/zero value for posts_per_page

### LINKED ISSUE #566: Divide by zero warning when posting a query

pwtyler:
Set posts_per_page to default to 1 if not set.

Replaces #567
Fixes #566.

### DIFF
--- a/includes/class-solrpower-wp-query.php
+++ b/includes/class-solrpower-wp-query.php
@@ -174,23 +174,27 @@ class SolrPower_WP_Query {
 	function posts_request( $request, $query ) {
 		if ( ! $this->is_solr_query( $query ) || false === SolrPower_Api::get_instance()->ping ) {
 			return $request;
-		}
-		add_filter( 'solr_query', array( SolrPower_Api::get_instance(), 'dismax_query' ), 10, 2 );
+    }
+
+    add_filter( 'solr_query', array( SolrPower_Api::get_instance(), 'dismax_query' ), 10, 2 );
+
 		$solr_options = SolrPower_Options::get_instance()->get_option();
-
 		$the_page = ( ! $query->get( 'paged' ) ) ? 1 : $query->get( 'paged' );
+    $qry = $this->build_query( $query );
 
-		$qry = $this->build_query( $query );
 		if ( '' === $qry ) { // If we don't have anything to query, let's do a wildcard.
 			$qry = '*';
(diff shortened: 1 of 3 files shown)


includes/class-solrpower-wp-query.php
```diff
@@ -175,22 +175,26 @@ function posts_request( $request, $query ) {
...
 		}
-		$this->qry = $qry;
-		$offset    = $query->get( 'posts_per_page' ) * ( $the_page - 1 );
-		$count     = $query->get( 'posts_per_page' );
-		$fq        = $this->parse_facets( $query );
-		$sortby    = ( isset( $solr_options['s4wp_default_sort'] ) && ! empty( $solr_options['s4wp_default_sort'] ) ) ? $solr_options['s4wp_default_sort'] : 'score';
-		$order     = ( $query->get( 'order', false ) ) ? strtolower( $query->get( 'order' ) ) : 'desc';
-		$extra     = array();
+
+		$posts_per_page = ( ! $query->get( 'posts_per_page' ) ) ? 1 : $query->get( 'posts_per_page' );
```
jazzsequence, line 189: I think this should be set to a more intelligent default if it's not set.

```suggestion
		$posts_per_page = get_option( 'posts_per_page' ) ?? 10;
		$posts_per_page = ( ! $query->get( 'posts_per_page' ) ) ? $posts_per_page : $query->get( 'posts_per_page' );
```

jazzsequence (changes requested):
use `posts_per_page` value from the database as a default rather than hard-coding to 1.

includes/class-solrpower-wp-query.php
pwtyler: Not sure if two nested defaults is worse than two ternaries but I didn't like ^ lol
```suggestion
		$posts_per_page = $query->get('posts_per_page', get_option('posts_per_page', 10));
```

jazzsequence (approved):

jazzsequence:
:shipit:

PR #603, opened 2023-07-26
Create release following tag

pwtyler:
Always be releasing.

On merge to main, build step will immediately publish a release, triggering the WordPress.org publish workflow.

### DIFF
--- a/.github/workflows/build-tag.yml
+++ b/.github/workflows/build-tag.yml
@@ -36,5 +36,7 @@ jobs:
         git commit -m "Release $VERSION"
         git tag "$VERSION"
         git push --tags
+        gh release create "$VERSION" -t "$VERSION" --generate-notes
       env:
         TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_TOKEN: ${{ github.token }}


.github/workflows/build-tag.yml
```diff
@@ -36,5 +36,7 @@ jobs:
         git commit -m "Release $VERSION"
         git tag "$VERSION"
         git push --tags
+        gh release create "$VERSION" -t "$VERSION" --generate-notes
       env:
-        TOKEN: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+        TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_TOKEN: ${{ github.token }}
```
jazzsequence, line 42: what's the GH_TOKEN variable for?

jazzsequence (approved):

.github/workflows/build-tag.yml
pwtyler: GH_TOKEN is what `gh` uses to auth access back to this repo (or any other). I am not actually sure if `TOKEN: ${{ secrets.GITHUB_TOKEN }}` is needed for the build step (I know its needed for the tests in the other file) but leaving as is for now.

PR #635, opened 2024-11-18
[CMSI 598] - Removing binding.pem from php container and using mtlsproxyhost for search-gateway {Wordpress Site}

amarnath-ayyadurai-23:
https://getpantheon.atlassian.net/browse/CMSI-598

dep: https://github.com/pantheon-systems/infrastructure/pull/3786


pwtyler:
As discussed in Slack, breaking changes that rely on running the latest version of the plugin are difficult.

- Env variables should be adjusted where they are set rather than adding new ones to the plugin
- use a self-generated cert at /certs/binding.pem that can't be used for mTLS.

We might consider making backwards compatible changes (i.e. wrap the `curl_setopt` in a `if ! file_exists`, or a function that checks multiple env variables) allowing us to clean up compatibility choices after a certain time frame or adoption threshold.

Issue #669, opened 2026-08-04
No CI coverage for the npm/grunt asset build


mehta-asim:
## Problem

No workflow in `.github/workflows/` runs `npm` or `grunt`. The asset build is entirely unexercised by CI, so a change that breaks CSS or JS generation can merge with every check green.

Verified: `grep -rnE "npm |grunt|node-version" .github/workflows/` returns nothing. `lint-test.yml` runs `composer install` only (lines 27 and 74).

## How it surfaced

#667 bumped `postcss` to 8.5.12 via the lockfile while leaving `grunt-autoprefixer` 3.0.4 and `autoprefixer-core` 5.2.1 in place. Both are written against the postcss 4 API. That PR showed 8 passing checks and was mergeable. Running the build locally on its branch:

```
$ npx grunt default
Loading "autoprefixer.js" tasks...ERROR
>> Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/vendor' is not defined by "exports" in node_modules/postcss/package.json
Warning: Task "autoprefixer" not found.
Aborted due to warnings.
```

Every green check passed while `grunt default` could not load a task in the `default` chain. #667 was closed in favour of #668, which migrates off the dead packages, but the coverage gap that hid the breakage remains.

The exposure is not theoretical for releases: `assets/css/*.css` is gitignored (`.gitignore:6`), so the shipped CSS is generated at release time from whatever the toolchain resolves then. A broken build is discovered while cutting a release rather than in review.

## Suggested fix

Add a job to `lint-test.yml` that installs node dependencies and runs the build, failing the check on a non-zero exit:

- `actions/setup-node` with the version the project targets
- `npm ci`
- `npx grunt default`

Optionally assert the expected artifacts exist afterward (`assets/css/admin.css`, `assets/css/facet.css`, and the `.min.css` pair) so a task that silently no-ops is caught too.

## Related

- #212 (2017) asks for `grunt jshint` to be split into its own build step. It presumes grunt ran in CI at the time. This issue is the broader precondition: grunt does not run in CI at all today, so #212 may want revisiting once a build job exists.
- #668 is the migration off `grunt-autoprefixer` and `autoprefixer-core` that this gap allowed to go unnoticed.

PR #658, opened 2026-07-27
fix(SITE-5995): resolve plugin-check compliance errors

mehta-asim:
## Summary

- Rename text domain `solr-for-wordpress-on-pantheon` → `solr-power` to match the plugin slug (renames the `.pot` accordingly)
- Add ABSPATH direct-access guards to entrypoint/template files
- Escape all output per context (`esc_html`/`esc_url`/`esc_attr`/`wp_kses_post`)
- Add `.distignore` so dev/test files are excluded from the dist archive (keeps `assets/` and `vendor/`, both required at runtime)
- Add `License` header; bump `Tested up to: 7.0`
- Replace `TEMPLATEPATH`, `mt_rand()`, `strip_tags()` with WP equivalents; fix deprecated `add_option()` 3rd param
- `phpcs:ignore` the low-level Solr cURL/stream client (direct control required; converting risks breaking Solr connectivity covered by Behat)

## Context

[SITE-5995](https://getpantheon.atlassian.net/browse/SITE-5995)

Resolves the `wordpress/plugin-check-action` compliance ERRORs surfaced by the migration in #657. Split from the migration PR to keep each reviewable.

## Test plan

- [ ] `Validate Plugin` check passes
- [ ] Behat suite passes (Solr connectivity intact after cURL client left as-is)
- [ ] Admin UI CSS/JS still load (assets/ retained in dist)

[SITE-5995]: https://getpantheon.atlassian.net/browse/SITE-5995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #657, opened 2026-07-22
ci(SITE-5995): add wporg-validator.yml using plugin-check-action

mehta-asim:
## Summary

- Adds `wporg-validator.yml` workflow using `wordpress/plugin-check-action@1.1.5` (SHA-pinned)
- Runs on `pull_request` and `push` to `main`
- Includes COMPOSER_AUTH fix (`env: COMPOSER_AUTH: '{"github-oauth": {}}'` + `rm -f ~/.composer/auth.json`) to avoid `shivammathur/setup-php` writing an invalid `ghs_*` token to `~/.composer/auth.json`
- Pins `wp-cli/dist-archive-command:v3.1.0`

## Context

[SITE-5995](https://getpantheon.atlassian.net/browse/SITE-5995)

`pantheon-systems/action-wporg-validator` is deprecated. solr-power never had any validator. This PR adds net-new plugin-check validation following the pattern established across other Pantheon WP plugins (wp-saml-auth, PAPC, wp-native-php-sessions).

A follow-up compliance PR will fix any ERRORs surfaced by the first CI run.

## Test plan

- [ ] CI runs and produces plugin-check output (pass or fail with specific error codes)
- [ ] No `COMPOSER_AUTH` errors in dist-archive step

[SITE-5995]: https://getpantheon.atlassian.net/browse/SITE-5995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #670, opened 2026-08-05
fix(deps): [VUL-26008 et al, SITE-5995] promote dev-dep security fixes + wporg-validator workflow to main [ci skip]

mehta-asim:
## Summary

Promotes to `main` the dev-dependency security fixes that have accumulated on `develop`, plus the SITE-5995 plugin-check workflow. Five files, `+445 / -326`, of which `package-lock.json` is `+693 / -...` regenerated noise.

`develop` is the working branch on this repo, but vulntools reads `main`, so those fixes are invisible to the scanners today. `main` is also the default branch and has had no wp.org plugin-check coverage at all.

Nothing here is plugin runtime code. No file under `includes/`, `template/`, `templates/`, `views/` or `solr-power.php` is touched, and no version string moves.

Per Phil in [the squad thread](https://pantheon.slack.com/archives/C0AD8TY61D0/p1785848771769979): since these are all dev deps, PR them to main with `[ci skip]` and reconcile back to develop, rather than cutting a release.

## What the diff actually contains

| File | Change | Ticket |
|---|---|---|
| `composer.json` | `+6` | VUL-24495, VUL-25207 |
| `package.json` | `+9 / -1` | VUL-24442, VUL-25001, VUL-25002, VUL-23981, VUL-26008 |
| `Gruntfile.js` | `+6 / -6` | VUL-26008 |
| `package-lock.json` | regenerated | all of the above |
| `.github/workflows/wporg-validator.yml` | `+50`, new file | SITE-5995 |

### `composer.json`

Adds a `config.audit.ignore` block and nothing else:

```json
"audit": {
    "ignore": [
        "GHSA-g446-98w2-8p5w",
        "GHSA-c2w2-prh8-qm98"
    ]
}
```

- `GHSA-g446-98w2-8p5w` is *Guzzle: Cookie Disclosure and Injection via IP-Address Domains* (`guzzlehttp/guzzle`).
- `GHSA-c2w2-prh8-qm98` is *guzzlehttp/psr7: Host Confusion via Weak URI Host Validation* (`guzzlehttp/psr7`).

Both reach this repo only through the dev toolchain. `require` is untouched and still resolves solarium/solarium and symfony/event-dispatcher.

### `package.json` and `Gruntfile.js`

`grunt-autoprefixer ~3.0.4` is dead upstream and pulls the vulnerable transitive tree. It is replaced with the maintained fork plus postcss 8:

- removed: `grunt-autoprefixer`
- added: `@lodder/grunt-postcss ^3.1.1`, `autoprefixer ^10.5.4`, `postcss ^8.5.23`
- added: `overrides: { "js-yaml": "3.15.0" }`, which is the js-yaml pin covering the four solr-power js-yaml tickets

`Gruntfile.js` follows the rename in four places: the task config block `autoprefixer:` becomes `postcss:`, `loadNpmTasks` switches to `@lodder/grunt-postcss`, and both the `watch` task list and the `default` registration swap `'autoprefixer'` for `'postcss'`.

One behavior detail worth checking on review: browser targeting moved rather than disappeared. The Gruntfile's `options.browsers = ['last 5 versions']` is gone, because grunt-postcss does not take it, and the same value now lives in `package.json` as a top-level `browserslist` key, which is where autoprefixer 10 reads it from. Same targets, different home.

### `package-lock.json`

Regenerated, not hand-edited. Resolved versions in the committed lockfile:

| Package | Resolved |
|---|---|
| `postcss` | 8.5.25 |
| `autoprefixer` | 10.5.4 |
| `@lodder/grunt-postcss` | 3.1.1 |
| `js-yaml` | 3.15.0 (single entry in the tree) |
| `grunt-autoprefixer` | absent |

Lockfile `version` stays `2.6.0`, matching `main`.

### `.github/workflows/wporg-validator.yml`

`167fa40` is a `git cherry-pick -x` of `dc70ec2` from `develop`, where it landed via #657. Byte-identical to develop's copy, verified with `git diff origin/develop HEAD -- .github/workflows/wporg-validator.yml` returning empty, so it will not conflict when `main` is reconciled back into `develop`.

The workflow builds a `wp dist-archive` on PHP 8.4 and runs `wordpress/plugin-check-action@6f5a57e` (1.1.5) against it. It triggers on `pull_request` and on `push` to `main`.

## `Validate Plugin` will be red here, and on `main` after merge

Expected, not a regression. `main` does not yet carry the SITE-5995 compliance fixes:

- `readme.txt` on `main` has `Tested up to: 6.7.1`, which trips `outdated_tested_upto_header`. That is an ERROR, not a warning.
- 12 `WordPress.Security.EscapeOutput.OutputNotEscaped` ERRORs across `template/s4w_search.php`, `template/s4wp_search.php`, `includes/legacy-functions.php` and `includes/class-solrpower-sync.php`. Those fixes are in #658, open against `develop` and green there.

`Validate Plugin` is **not** a required check on `main`. The ruleset requires only `Wiz IaC Scanner`, `Wiz Vulnerability Scanner`, `Wiz Data Scanner` and two approvals, so the red does not block this merge. It goes green once #658 lands on `develop` and `develop` is promoted to `main`.

## Deliberately left on `develop`

- `ea045c3` bumps 2.6.0 to 2.6.1-dev across `solr-power.php`, `readme.txt`, `README.md`, `CHANGELOG.md` and `package.json`. Prod-visible version string, implies a release.
- `d8893c3` new wp.org listing imagery.

## Why `[ci skip]` is in the title and not the commits

`.github/workflows/build-tag.yml` triggers on `push` to `main` and cuts a tag and release. GitHub decides whether to skip based on the **HEAD commit message of the push**, which for a squash merge is the squash commit message, defaulting to the PR title.

So `[ci skip]` belongs in the PR title. Putting it in the branch commits instead would also skip this PR's own validation, which we want to run.

**When merging: squash, and keep `[ci skip]` in the squash commit message.** If the repo is set to a merge commit instead, put `[ci skip]` in the merge commit message.

Safety net: `build-tag-release@v0` runs with `draft: "true"`, so even if the skip does not take effect the release is created as a draft rather than published.

## Verification

- `grunt default` run locally, passes end to end including the `postcss:files` task, producing both stylesheets and their sourcemaps. This matters because the repo has **no CI coverage for the npm/grunt build** (tracked in #669), so CI cannot catch a broken build here.
- `composer validate` reports `./composer.json is valid`, no lockfile staleness.
- Resolved lockfile versions confirmed directly against the committed `package-lock.json`, table above.
- Both GHSA IDs resolved against the GitHub Advisory API to confirm they are the guzzle and psr7 advisories, not something else.
- `wporg-validator.yml` confirmed byte-identical to develop's copy.

## Follow-up

- Reconcile `main` back into `develop` after this merges, so the two do not diverge further.
- Merge #658 to `develop`, then promote `develop` to `main`, which turns `Validate Plugin` green on `main`.
- VUL-25911 (wpcs) is **not** addressed here. Blocked by `config.platform.php` pinned to 7.1 while wpcs 3.4.1 requires `php >=7.2`, and needs a product decision on the plugin's minimum PHP. Tracked separately.

## Test plan

- [ ] Wiz scanners green
- [ ] `Validate Plugin` reports only the known `main` ERRORs listed above, nothing beyond them
- [ ] Merge as a squash with `[ci skip]` retained in the commit message
- [ ] Confirm `Build, Tag, and Release` did not run on the resulting push to `main`
- [ ] Confirm no new tag or release was created
- [ ] Reconcile `main` back to `develop`

### DIFF
--- a/.github/workflows/wporg-validator.yml
+++ b/.github/workflows/wporg-validator.yml
@@ -1,0 +1,50 @@
+name: WP.org Validator
+on:
+  pull_request:
+  push:
+    branches:
+    - main
+
+permissions:
+  contents: read
+
+jobs:
+  validate:
+    name: Validate Plugin
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+    - name: Setup PHP 8.4
+      uses: shivammathur/setup-php@44454db4f0199b8b9685a5d763dc37cbf79108e1 # v2
+      with:
+        php-version: "8.4"
+        extensions: mysqli
+        tools: composer,wp-cli
+
+    - name: Cache Composer dependencies
+      uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+      with:
+        path: |
+            ~/.composer/cache
+            vendor
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+            ${{ runner.os }}-composer-
+
+    - name: Install dependencies
+      run: composer install --no-dev --no-interaction --no-progress --prefer-dist --no-scripts --no-ansi
+
+    - name: Create Dist Archive for wp.org validation
+      env:
+        COMPOSER_AUTH: '{"github-oauth": {}}'
+      run: |
+        rm -f ~/.composer/auth.json
+        wp package install wp-cli/dist-archive-command:v3.1.0
+        wp dist-archive . --plugin-dirname=solr-power --filename-format=solr-power-dist
+        unzip ../solr-power-dist.zip
+
+    - name: Run plugin check
+      uses: wordpress/plugin-check-action@6f5a57e173c065a394b78688f75df543e4011902 # 1.1.5
+      with:
+        build-dir: solr-power
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -63,11 +63,11 @@ module.exports = function( grunt ) {
       /**
        * Auto-prefix CSS Elements after SASS is processed.
        */
-      autoprefixer : {
+      postcss : {
 
           options : {
-              browsers : ['last 5 versions'],
-              map      : true
+              map        : true,
+              processors : [ require( 'autoprefixer' )() ]
           },
 
           files : {
@@ -146,7 +146,7 @@ module.exports = function( grunt ) {
                   'assets/css/scss/*'
               ],
 
-              tasks : ['sass', 'autoprefixer', 'cssmin']
+              tasks : ['sass', 'postcss', 'cssmin']
 
           }
       },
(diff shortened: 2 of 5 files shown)


tundeafolabi (approved):
lgtm

pwtyler (approved):

